PR #8, opened 2018-06-07
SingleSignOn

tobiasKaminsky:
Thanks to the great work by @David-Development, we will soon have a SingleSignOn.
I just created the PR from his branch, so that code review is easier.


src/main/java/de/luhmer/owncloud/accountimporter/adapter/AccountImporterAdapter.java
```diff
@@ -43,28 +43,26 @@ public View getView(final int position, View view, ViewGroup parent) {
...
             holder = new ViewHolder(text1, text2, cbChecked);
-
             view.setTag(holder);
         }
 
-        holder.text1.setText(getItem(position).type);
-        holder.text2.setText(getItem(position).url);
-        holder.cbChecked.setChecked(getItem(position).checked);
+        String username = getItem(position).name.split("@")[0];
+        String server   = getItem(position).name.split("@")[1];
```
tobiasKaminsky: Newest accounts should have userId@server, instead of loginname@server.
UserId is a unique id per server
Previously we used the loginname, which also can be an email address.

To also have this working with legacy accounts, this should be splitted be last "@":
tobias@mail.com@nextcloud.cloud.com -> tobias@mail.com (username) & nextcloud.cloud.com (server)

David-Development: Preview of code: https://repl.it/repls/KookyTriangularInterchangeability

Better ideas on how to implement this? Unfortunately Java does not offer something like "split last"..

mario:
Two things to discuss from the Nextcloud files side, to make this REALLY perfect:

- move away from error-prone account manager to internal database
- start storing server protocol, avoiding costly and useless network calls

build.gradle
```diff
@@ -21,7 +31,19 @@ android {
...
+    // https://mvnrepository.com/artifact/com.google.code.gson/gson
+    compile group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
+
+    compile 'io.reactivex.rxjava2:rxandroid:2.0.1'
+    // Because RxAndroid releases are few and far between, it is recommended you also
+    // explicitly depend on RxJava's latest version for bug fixes and new features.
+    compile 'io.reactivex.rxjava2:rxjava:2.1.0'
+
+    compile group: 'commons-io', name: 'commons-io', version: '2.5'
 }
```
mario: New line :P

src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,15 @@
...
-    </application>
+    <uses-permission android:name="android.permission.GET_ACCOUNTS" />
+    <uses-permission
+        android:name="android.permission.USE_CREDENTIALS"
+        android:maxSdkVersion="22" />
 
-</manifest>
+    <application android:label="@string/app_name"></application>
+
+</manifest>
```
mario: New line.

src/main/aidl/de/luhmer/owncloud/accountimporter/helper/IInputStreamService.aidl
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package de.luhmer.owncloud.accountimporter.helper;
+
+// Declare the interface.
+interface IInputStreamService {
+
+    ParcelFileDescriptor performNextcloudRequest(in ParcelFileDescriptor input);
+
+}
```
mario: New line.

src/main/aidl/de/luhmer/owncloud/accountimporter/helper/IInputStreamService.aidl
```diff
@@ -0,0 +1,25 @@
...
+	From _The Busy Coder's Guide to Advanced Android Development_
+		http://commonsware.com/AdvAndroid
+
+
+	More information here: https://github.com/abeluck/android-streams-ipc
+*/
+
+package de.luhmer.owncloud.accountimporter.helper;
+
+// Declare the interface.
```
mario: Self evident.

src/main/java/de/luhmer/owncloud/accountimporter/helper/AccountImporter.java
```diff
@@ -2,153 +2,147 @@
...
+        List<Account> accountsAvailable = new ArrayList<>();
         for (Account account : accounts) {
-            String aType = account.type.intern();
-
-            if (//aType.equals("org.dmfs.caldav.account") ||
-                // aType.equals("org.dmfs.carddav.account") ||
-                    aType.equals("de.luhmer.tasksync") ||
-                            aType.equals("owncloud")) {
-                //accountsAvailable.add(accounts[index].type);
+            if (account.type.equals("nextcloud")) {
```
mario: This  "nextcloud" thing needs to be configurable.

src/main/java/de/luhmer/owncloud/accountimporter/helper/NextcloudAPI.java
```diff
@@ -0,0 +1,238 @@
...
+        if (type != Void.class) {
+            result = gson.fromJson(targetReader, type);
+            if (result != null) {
+                Log.d(TAG, result.toString());
+            }
+        }
+        targetReader.close();
+
+        /*
+        byte[] header = new byte[1];
```
mario: Is any of this needed? If not, remove.

src/main/java/de/luhmer/owncloud/accountimporter/helper/NextcloudRequest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        public Builder setAccountName(String accountName) {
+            ncr.accountName = accountName;
+            return this;
+        }
+    }
+
+
+
+}
```
mario: New line.

src/main/res/layout/import_accounts_fragment.xml
```diff
@@ -1,23 +1,23 @@
...
-        android:layout_height="match_parent"
-        android:layout_gravity="center"/>
+        android:layout_width="50dp"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:layout_centerHorizontal="true"
+        android:layout_centerVertical="true"/>
 
-        -->
 </LinearLayout>
```
mario: New line.

mario:
@David-Development can we please get copyright headers at top of each file?

tobiasKaminsky:
> move away from error-prone account manager to internal database

I think we should postpone this as this will heavily affect the Files app and thus should be tested & discussed in detail.

> start storing server protocol

This is already marked as todo

> avoiding costly and useless network calls

What do you mean with that? This library is acting as a "bridge" between calling package and files app/ server, so we have no influence on which calls are made?

tobiasKaminsky:
@David-Development do you have the time to make the changes or may I jump in?

mario:
@tobiasKaminsky  I mean: because Files app doesn't store the protocol, we'll need to do a check in every app if it's https or http. Storing the protocol would help in this.

build.gradle
David-Development: Where exactly? Or do you mean that there shouldn't be any empty lines in the dependency section? :D

src/main/java/de/luhmer/owncloud/accountimporter/helper/AccountImporter.java
David-Development: Why? Aren't all nextcloud accounts stored with the type "nextcloud"?

build.gradle
tobiasKaminsky: The file should end with a new line.
It is some kind of posix standard, which I am not sure if we really need it (and thus forget it very often), though it cannot be bad, to have it ;-)

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206 
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

src/main/java/de/luhmer/owncloud/accountimporter/helper/AccountImporter.java
tobiasKaminsky: The files app is brandable, so e.g. if you want to change everything from Nextcloud to Ubercloud, also the account type is "ubercloud" as otherwise you would see the other accounts in your "new" app.

David-Development:
I removed a lot of code that is not needed anymore. (There were some leftovers from some the early prototypes that I build..). And everything you guys mentioned is fixed now.

src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,14 @@
...
 
-    </application>
+    <uses-permission android:name="android.permission.GET_ACCOUNTS" />
+    <uses-permission
+        android:name="android.permission.USE_CREDENTIALS"
+        android:maxSdkVersion="22" />
 
-</manifest>
+    <application android:label="@string/app_name" />
+</manifest>
```
mario: New line missing.

src/main/aidl/de/luhmer/owncloud/accountimporter/helper/IInputStreamService.aidl
```diff
@@ -0,0 +1,22 @@
...
+
+
+	More information here: https://github.com/abeluck/android-streams-ipc
+*/
+
+package de.luhmer.owncloud.accountimporter.helper;
+
+interface IInputStreamService {
+    ParcelFileDescriptor performNextcloudRequest(in ParcelFileDescriptor input);
+}
```
mario: New line.

src/main/java/de/luhmer/owncloud/accountimporter/helper/AccountImporter.java
David-Development: Oh okay, I see. Any good ideas on how to change that accordingly? Another question would be.. how do we know which "brand" the user wants to use? 

Apart from that, I don't think that we need to check the type in the future since starting with Android 8,  the `accMgr.getAccounts();` returns only accounts that the app has access to. (https://developer.android.com/about/versions/oreo/android-8.0-changes#o-sec)

The client is basically the only app that needs to know about the account type then.. as seen here: https://github.com/nextcloud/news-android/blob/sso/News-Android-App/src/main/java/de/luhmer/owncloudnewsreader/LoginDialogFragment.java#L249

mario:
E.g. in Talk I show the server url in the profile. I'd like to continue doing so.

tobiasKaminsky:
Ok, passing the server url as info is of course possible.

I would like to merge this now and then refine the whole library with smaller PRs rather than trying to solve everything right here.

So my roadmap would be
- merge this
- tomorrow: 
	- open up a new PR (here & change files PR) for passing a token between 3rd party app and NC files app, instead of a real auth token from NC (as described here https://github.com/nextcloud/Android-SingleSignOn/issues/9)
	- adjust News app to use library snapshot via jitpack

tobiasKaminsky:
@David-Development 
I updated this to also use calling package name. We have a hardcoded list of package names, which are allowed (talked to @rullzer about this).
I also updated https://github.com/nextcloud/news-android/pull/639, so you can test it

src/main/java/de/luhmer/owncloud/accountimporter/helper/AccountImporter.java
rullzer: For now this is not a big deal. Anything branded will just be files+talk I guess.

David-Development:
Just tested it and it works great! Thank you! 👍 And I totally agree with what you said in #11 and #12. This library does not support it yet either.. (see https://github.com/nextcloud/Android-SingleSignOn/blob/sso/src/main/java/de/luhmer/owncloud/accountimporter/helper/AccountImporter.java#L55)

But I can't think of a clean/good solution for multi-account support yet.

David-Development:
Right now I can think of two use cases when a user wants to use multiple accounts:

1. Just as in the Files App, the user can switch between different accounts, but only one account is active at all times. This is very "light" version of multi-account usage. I think this is already achieved since the client app can use the `FindAccounts` method to get all accounts and then select one with the `SetCurrentAccount` method. 

2. Just as I'm using it in my task-sync app, I let the user add multiple calendars from multiple caldav endpoints / servers. Therefore all accounts will be accessed somewhat simultaneously. Therefore we can't just use the `SetCurrentAccount` method to switch between the accounts. We need to hold a reference to the accounts. Not sure if this should be part of this sso library or if it's the responsibility of the client app itself. Any thoughts on this?

Another problem, which also affects case 1 is, that the `FindAccounts` will only return accounts that the user have granted access to (due to the new security restrictions on Android 8+)

tobiasKaminsky:
Merged, as it is working.
@David-Development regarding #11 and #12, we should discuss this is in the corresponding issues.

PR #10, opened 2018-06-08
Change build.gradle

tobiasKaminsky:
- add gradle wrapper


tobiasKaminsky:
This then allows jitpack to build snapshots or, later, releases.

PR #18, opened 2018-06-18
Add retrofit helper

### LINKED ISSUE #17: Add retrofit helper/wrapper functions
Another question is, if we should add the following class, which provides some helpful retrofit/rxjava methods to this library. I think we should, but in case we add it, we have to add the retrofit/rxjava dependencies to this library as well. But I think most apps use retrofit anyways.. right? Especially with this library it's really easy to use..

https://github.com/nextcloud/news-android/blob/master/News-Android-App/src/main/java/de/luhmer/owncloudnewsreader/reader/nextcloud/API_SSO_Helper.java

tobiasKaminsky:
Fix #17


tobiasKaminsky:
@David-Development feel free to move the files around :-D

src/main/java/de/luhmer/owncloud/accountimporter/helper/API_SSO_Helper.java
```diff
@@ -0,0 +1,132 @@
...
+import retrofit2.Response;
+
+public class API_SSO_Helper {
+
+    public static ResponseBody getResponseBodyFromRequest(NextcloudAPI nextcloudAPI, NextcloudRequest request) {
+        try {
+            InputStream os = nextcloudAPI.performNetworkRequest(request);
+            return ResponseBody.create(null, 0, new BufferedSourceSSO(os));
+        } catch (Exception e) {
+            e.printStackTrace();
```
mario: Let's avoid stacktraces.

mario: Everywhere :)

src/main/java/de/luhmer/owncloud/accountimporter/helper/BufferedSourceSSO.java
```diff
@@ -0,0 +1,249 @@
...
+    @Override
+    public Timeout timeout() {
+        return null;
+    }
+
+    @Override
+    public void close() throws IOException {
+
+    }
+}
```
mario: New line.

src/main/java/de/luhmer/owncloud/accountimporter/helper/API_SSO_Helper.java
```diff
@@ -0,0 +1,132 @@
...
+
+import io.reactivex.Completable;
+import io.reactivex.functions.Action;
+import okhttp3.Request;
+import okhttp3.ResponseBody;
+import retrofit2.Call;
+import retrofit2.Callback;
+import retrofit2.Response;
+
+public class API_SSO_Helper {
```
mario: Hm, why classes names with underscores? o.O

David-Development: I'll rename and move this files :D It was just a quick proof of concept

David-Development:
@mario @tobiasKaminsky what should we do with all the methods, that are not "implemented" yet? Right now they are all there but they might return some "false" values. Such as if you make a request and you ask "executed" it'll always return false.. 

Maybe we should throw an `new UnsupportedOperationException("")` so that developers know, that this isn't implemented yet. (They can feel free to add it to this library then)

David-Development:
Okay, just pushed a bunch of changes and refactorings :)

tobiasKaminsky:
> Maybe we should throw an new UnsupportedOperationException("")

Fine by me :+1:

tobiasKaminsky:
@David-Development some minor (project related) notes
- on files repo we decided to do a rebase prior merging, so that all the old merge commits get eliminated
- I saw that you pushed some changes directly (not related to this PR). To have all contributors aware of the newest change, it would be cool that you open up a new PR instead of directly pushing it.

David-Development:
Oh okay, sorry! I never used rebase before but I'll try to use it from now on. 

You mean that I added the Exceptions? `new UnsupportedOperationException("")`

tobiasKaminsky:
:-D
No, the exceptions are fine.

I was referring to https://github.com/nextcloud/Android-SingleSignOn/commit/70831be877652f096b4998712ce83f5664a76b98, which changed the GetCurrentAccount and while adjusting News-App I stumbled about this, as I did not remembered any changes there ;-)

David-Development:
Ohhh okay, I see. So you wanted me to open a PR instead of pushing it to the master branch right away? Makes sense! I'll create a PR next time! :)

tobiasKaminsky:
Is this fine from your side, David?

David-Development:
Yes, it is! :)

PR #21, opened 2018-06-18
Refactoring / Update Readme

David-Development:
We need to merge #18 first since the changes made here are based on the `addRetrofitHelper` branch.

Changes: 
- remove unused interface
- make AlertDialog "clickable" when files app is not installed (error message) (to open Play-Store)
- update examples to match latest api specifications / include more useful example api calls


tobiasKaminsky:
On files app, we launch installation directly:
https://github.com/nextcloud/android/blob/82679c70f287ce337539e82a948945adecf1f26c/src/main/java/com/owncloud/android/ui/activity/Preferences.java#L811-L824

First play store is tried, then fdroid.
Do we want to do it the same?

PR #33, opened 2018-09-28
remove application tag

Unpublished:
This is not an application and it overwrites app_name of https://github.com/nextcloud/news-android (at least with Android Studio 3.2)


David-Development:
Even after upgrading to Android Studio 3.2 the news app is still named "News" on my Android 7. But I totally agree, that this tag should be removed as it is not needed.

Thank you for your contribution! 👍

PR #44, opened 2019-02-18
create account object when account list is empty

PhieF:
Hi,

On android 7 nextcloud crashes after getting permission because AccountManager.getAccounts() returns an empty list on the SSO library

this is a quick fix (it works) but perhaps a bad one

what do you think ?

Thanks !

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/AccountImporter.java
+++ b/src/main/java/com/nextcloud/android/sso/AccountImporter.java
@@ -110,7 +110,7 @@ public class AccountImporter {
                 return account;
             }
         }
-        return null;
+        return new Account(name,"nextcloud");
     }
 
     public static void clearAllAuthTokens(Context context) {


tobiasKaminsky:
When and which app is crashing?

PhieF:
I'm using SSO on my app
When connecting to a nextcloud account, after authorizing the account, the nextcloud app crashes with this error

https://github.com/PhieF/CarnetDocumentation/issues/29#issuecomment-464362337

it is because AccountImporter.getAccountForName on the sso lib returns null, 

in findAccounts()
```
final AccountManager accMgr = AccountManager.get(context);
 final Account[] accounts = accMgr.getAccounts();
```
accounts is empty

perhaps it is a problem with my app, I don't know

tobiasKaminsky:
Is this already in your master branch? Then I can have a look.

PhieF:
On my app or on the SSO ? 

My app is here:

https://github.com/PhieF/Sync/blob/master/src/main/java/com/spisoft/sync/wrappers/nextcloud/NextCloudAuthorizeFragment.java

it crashes in
```
AccountImporter.onActivityResult

```

David-Development:
@tobiasKaminsky Did you have time to look into it yet?

tobiasKaminsky:
Hm. I tried it with lastest files / sso master and your app (also master) and it works for me.

Can you give us a more detailed description of your steps?

David-Development:
This is properly fixed in #62

Issue #12, opened 2018-06-12
Handle re-installation of 3rd party app


tobiasKaminsky:
- install 3rd party app
- allow access
- re-install 3rd party app
- token is already stored in files app (as it is based on package name), but 3rd party app has no longer the token
--> we need some way to re-authenticate.

tobiasKaminsky:
- [ ] remove 3rd party app after everything worked & re-install it --> should ask again for authentication
- [ ] remove files app after everything worked & re-install it -> 3rd party also should again ask for authentication

tobiasKaminsky:
First one works out of the box as we overwrite token on "grant access".

tobiasKaminsky:
Second one returns "invalid token", which is in https://github.com/nextcloud/android/blob/c52f9c842f44b5c9e42b1125468fa03dd554d162/src/main/java/com/nextcloud/android/sso/Constants.java#L17

@stefan-niedermann @desperateCoder 3rd party apps should check for this and start their own re-authenticate process again.

Would that work for you?

desperateCoder:
We could do it like this, but should we? This is a case, future users of the SSO have to be aware of. Is it impossible for the SSO lib to handle this?

For me this shouldn't be a problem, but it's inconvenient tbh.

tobiasKaminsky:
hm, I have no idea how to handle this via SSO:
- install files app & authenticate account
- install 3rd party app
- grant account to 3rd party
	- token is stored in files app (this is needed to verify app calls)
	- sso account is stored via shared preferences in 3rd party app
- remove files app --> no tokens are available
- re-install files app
- login into files app with account from above
- 3rd party app tries to access said account
--> files app will refuse as there is no token in files app

tl;dr: if files app is deleted, we do not know anything about already authenticated accounts/apps.

desperateCoder:
So how can you tell, when to throw the exception mentioned above? Isn't that exactly the case when we have to reauthentificate?

tobiasKaminsky:
The exception is thrown by SSO/files.
I'll change this in above scenario.

desperateCoder:
So now I'm confused. File app says CE_1. SSO still knows selected account. Can't SSO catch CE_1 and try to reauthentificate/re-grant?

Sorry if I am talking nonsense, I don't have had a deep-dive into the whole process.

AndyScherzinger:
[...]

PR #47, opened 2019-03-14
Fix for deck

stefan-niedermann:
cc @tobiasKaminsky @David-Development 

Please review and check especially whether or not the `clearAllAuthTokens` within the `pickNewAccount` is needed.

Thx!

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/AccountImporter.java
+++ b/src/main/java/com/nextcloud/android/sso/AccountImporter.java
@@ -68,7 +68,7 @@ public class AccountImporter {
         if (appInstalledOrNot(fragment.getContext(), "com.nextcloud.client")) {
 
             // Clear all tokens first to prevent some caching issues..
-            clearAllAuthTokens(fragment.getContext());
+            //clearAllAuthTokens(fragment.getContext());
 
             Intent intent = AccountManager.newChooseAccountIntent(null, null, new String[]{"nextcloud"},
                     true, null, null, null, null);
--- a/src/main/java/com/nextcloud/android/sso/helper/SingleAccountHelper.java
+++ b/src/main/java/com/nextcloud/android/sso/helper/SingleAccountHelper.java
@@ -29,14 +29,14 @@ import com.nextcloud.android.sso.model.SingleSignOnAccount;
 
 public final class SingleAccountHelper {
 
-    private static final String PREF_SINGLE_ACCOUNT_STRING = "PREF_ACCOUNT_STRING";
+    private static final String PREF_CURRENT_ACCOUNT_STRING = "PREF_CURRENT_ACCOUNT_STRING";
 
     private SingleAccountHelper() {
     }
 
     private static String getCurrentAccountName(Context context) throws NoCurrentAccountSelectedException {
         SharedPreferences mPrefs = AccountImporter.getSharedPreferences(context);
-        String accountName = mPrefs.getString(PREF_SINGLE_ACCOUNT_STRING, null);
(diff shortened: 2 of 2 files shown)


David-Development:
@stefan-niedermann Maybe you can elaborate a little why you removed the `clearAllAuthTokens` call? Did you run into some issues?

stefan-niedermann:
Yes, basically everytime i add a new Account on the deck app, it will `clearAllAuthTokens` - Imagine this workflow:

1. Open Deck 1st time
2. Select initial user A
3. Add second user B
4. Switch between users

So as soon as i will do Step 3, user A will be deleted from the shared preferences, therefore i can no longer switch between the accounts with `NextcloudFilesAppAccountNotFoundException`.

Maybe @tobiasKaminsky is able to explain it better ^^

David-Development:
@stefan-niedermann Got you.. I see. This is indeed a problem.

I'm trying to remember why we added it in the first place. Maybe #12 is related to this? Maybe @tobiasKaminsky knows more? 

Is there a case in which it is problematic to keep the tokens?

tobiasKaminsky:
> I'm trying to remember why we added it in the first place. Maybe #12 is related to this? Maybe @tobiasKaminsky knows more?

I also thought about it… 
I think #12 is really the root cause of this.
We then should have this handled differently…
Two ideas:
- upon uninstallation the 3rd party app needs to unregister via SSO lib -> will be not reliable
- upon re-installation we know on files app that we have a non-matching account and we can present a dialog to re-authenticate (my preferred solution)

David-Development:
When reinstalling an app the client app doesn't know anything about previous accounts, right? So I guess we can safely assume that in order to use the single sign on feature, it has to run the request authentication flow again which generates a new token. So we should be fine here. 

For reference: Permission Handling in the files app:
https://github.com/nextcloud/android/blob/086492b6d4f3514a84631d7c2539c0480438185c/src/main/java/com/owncloud/android/ui/activity/SsoGrantPermissionActivity.java#L55

The token is validated here:
https://github.com/nextcloud/android/blob/086492b6d4f3514a84631d7c2539c0480438185c/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java#L220

Just wondering.. It looks like were doing it on a per package basis, correct? Doesn't that mean that once the user grants permission to one account, he'll have access to all accounts? Assumed that he knows the name of the other accounts.

tobiasKaminsky:
I think we can merge this.

On a new PR I'll add/test this
- remove 3rd party app after everything worked & re-install it --> should ask again for authentication
- remove files app after everything worked & re-install it -> 3rd party also should again ask for authentication

On another one
- fix SSO to use triple of [account, package id, token] to authenticate.

David-Development:
@stefan-niedermann @desperateCoder @tobiasKaminsky 
Just added two methods for handling TokenMismatchExceptions. Please note that we cannot simply recover from an TokenMismatch all the time as in Android 9 we need to explicitly grant permission to an account before we can request the token. 

https://github.com/nextcloud/Android-SingleSignOn/blob/master/src/main/java/com/nextcloud/android/sso/AccountImporter.java#L226

And in case you already use the SingleAccountHelper:
https://github.com/nextcloud/Android-SingleSignOn/blob/master/src/main/java/com/nextcloud/android/sso/helper/SingleAccountHelper.java#L60


Example usage: https://github.com/nextcloud/news-android/blob/master/News-Android-App/src/main/java/de/luhmer/owncloudnewsreader/NewsReaderListActivity.java#L438

tobiasKaminsky:
Niceeeeeeee

PR #50, opened 2019-03-19
[WIP] add change listener

tobiasKaminsky:


David-Development:
@tobiasKaminsky Just out of curiosity, if there is a subscribe method.. isn't there an unsubscribe method?

https://developer.android.com/reference/android/content/SharedPreferences.html#unregisterOnSharedPreferenceChangeListener(android.content.SharedPreferences.OnSharedPreferenceChangeListener)

Please also note the warnings given in the docs: https://developer.android.com/reference/android/content/SharedPreferences.html#registerOnSharedPreferenceChangeListener(android.content.SharedPreferences.OnSharedPreferenceChangeListener)

tobiasKaminsky:
> Caution: The preference manager does not currently store a strong reference to the listener. You must store a strong reference to the listener, or it will be susceptible to garbage collection. We recommend you keep a reference to the listener in the instance data of an object that will exist as long as you need the listener.

I read this as it is enough that the listener exists all the time, for which 3rd party apps would be responsible.
Or is this not enough?
Then I could add this to a static field in SingleAccountHelper.

David-Development:
@tobiasKaminsky I think it should be the responsibility of 3rd party apps. However for the sake of completeness I think it would be good to have the unsubscribe method as well.

tobiasKaminsky:
I already force pushed the unregister method.

stefan-niedermann:
Sorry guys, i think you got me wrong... Now we have the same situation as if we would completely handle the shared preferences listener in the 3rd party apps. Simply because **we do not know the key**.

The register and unregister methods should not take a listener instance but create its own instance by using the PREF_CURRENT_ACCOUNT_STRING constant which **we do not know** in 3rd party clients.

tobiasKaminsky:
```
public static SharedPreferences.OnSharedPreferenceChangeListener registerOnAccountChangeListener(Context context) {
        SharedPreferences sharedPreferences = AccountImporter.getSharedPreferences(context);
        
        return new SharedPreferences.OnSharedPreferenceChangeListener() {
            @Override
            public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
                if (PREF_CURRENT_ACCOUNT_STRING.equals(key)) {
                    // do what?
                }
            }
        }
    }
```

Something like that? 
But how does the 3rd party app can then use it?

stefan-niedermann:
Maybe by passing an Consumer as second parameter (instead of the listener) which can be called on `// do what?`?

stefan-niedermann:
Alternatively: *if* we can read `PREF_CURRENT_ACCOUNT_STRING` then we can handle alle the listener stuff in 3rd-party apps as well, but i think a 3rd party app is not able to read it, is it?

tobiasKaminsky:
Hm. Consumer might work, but then we need the change listener to be somehow persistent, as mentioned in the docs.

I rather would not want to have PREF_CURRENT_ACCOUNT_STRING to be public / used by other apps. Knowledge of internal stuff should be as little as possible.

PR #52, opened 2019-04-05
implement InterceptorPattern for NetworkRequest

David-Development:
Allows devs to write unit tests by faking the server response. 

Example: (from Nextcloud News App)
```java
NextcloudAPI nextcloudAPI = new NextcloudAPI(GsonConfig.GetGson(), new NewsTestNetworkRequest(callback));

class NewsTestNetworkRequest extends NetworkRequest {

    NewsTestNetworkRequest(NextcloudAPI.ApiConnectedListener callback) {
        super(null, null, callback);
    }

    @Override
    protected InputStream performNetworkRequest(NextcloudRequest request, InputStream requestBodyInputStream) throws Exception {
        if(Looper.myLooper() == Looper.getMainLooper()) {
            throw new NetworkOnMainThreadException();
        }

        Log.e(TAG, "Requested URL: " + request.getUrl());
        InputStream inputStream = null;
        switch (request.getUrl()) {
            case "/index.php/apps/news/api/v1-2/feeds":
                inputStream = handleCreateFeed(request);
                break;
        }
        return inputStream;
    }


    private final String NEW_FEED_SUCCESS  = "http://test.de/new";
    private final String NEW_FEED_FAIL     = "http://test.de/fail";
    private final String NEW_FEED_EXISTING_ERROR_MESSAGE = "{\"message\":\"Feed konnte nicht hinzugef\\u00fcgt werden:  Existiert bereits\"}";
    private final String NEW_FEED_FAIL_ERROR_MESSAGE     = "{\"message\":\"FeedIo\\\\Adapter\\\\NotFoundException: Client error: `GET http:\\/\\/feeds2.feedburner.com\\/stadt-bremerhaven\\/dqXM222` resulted in a `404 Feed not found error: FeedBurner cannot locate this feed URI.`";

    // https://github.com/nextcloud/news/blob/master/docs/externalapi/Legacy.md#create-a-feed
    private InputStream handleCreateFeed(NextcloudRequest request) throws NextcloudHttpRequestFailedException {
        switch (request.getParameter().get("url")) {
            case NEW_FEED_SUCCESS:
                return stringToInputStream("");
            case NEW_FEED_FAIL:
                throw new NextcloudHttpRequestFailedException(422, new Throwable(NEW_FEED_FAIL_ERROR_MESSAGE));
        }
        return null;
    }

    private InputStream stringToInputStream(String data) {
        return new ByteArrayInputStream(data.getBytes(Charset.forName("UTF-8")));
    }
}

```

You might need those in your proguard config:
```bash
# https://stackoverflow.com/a/39777485
# Also, note that this rule should be added to the regular proguard file(the one of listed in proguardFiles) and not the test one(declared as testProguardFile)
# java.lang.NoSuchMethodError: No virtual method getParameter
-keepclasseswithmembers public class com.nextcloud.android.sso.aidl.NextcloudRequest { *; }
 No newline at end of file
```

Full Example:
https://github.com/nextcloud/news-android/blob/d94df30ddfe987cf5f2482ac51aa1cf0b6e3beec/News-Android-App/src/androidTest/java/de/luhmer/owncloudnewsreader/di/TestApiProvider.java

https://github.com/nextcloud/news-android/blob/d94df30ddfe987cf5f2482ac51aa1cf0b6e3beec/News-Android-App/src/androidTest/java/com/nextcloud/android/sso/api/NewFeedTests.java

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ android {
 
 dependencies {
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
-    implementation "androidx.appcompat:appcompat:1.1.0-alpha04"
+    implementation "androidx.appcompat:appcompat:1.0.2"
 
     implementation 'com.google.code.gson:gson:2.8.5'
 
@@ -57,5 +57,5 @@ dependencies {
     // Required for local unit tests (JUnit 4 framework)
     testImplementation 'junit:junit:4.12'
     // required if you want to use Mockito for unit tests
-    testImplementation 'org.mockito:mockito-core:2.25.1'
+    testImplementation 'org.mockito:mockito-core:2.26.0'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,7 @@ android {
 
 dependencies {
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
+    implementation "androidx.appcompat:appcompat:1.1.0-alpha04"
 
     implementation 'com.google.code.gson:gson:2.8.5'
 
--- a/src/main/java/com/nextcloud/android/sso/AccountImporter.java
+++ b/src/main/java/com/nextcloud/android/sso/AccountImporter.java
@@ -61,6 +61,8 @@ public class AccountImporter {
     public static final int CHOOSE_ACCOUNT_SSO = 4242;
     public static final int REQUEST_AUTH_TOKEN_SSO = 4243;
 
+    private static SharedPreferences SHARED_PREFERENCES;
+
     public static boolean accountsToImportAvailable(Context context) {
         return findAccounts(context).size() > 0;
     }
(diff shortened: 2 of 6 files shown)


David-Development:
@desperateCoder @stefan-niedermann @AndyScherzinger maybe this is useful for you guys as well?

build.gradle
```diff
@@ -40,6 +40,7 @@ android {
 
 dependencies {
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
+    implementation "androidx.appcompat:appcompat:1.1.0-alpha04"
```
AndyScherzinger: Maybe rather go for a 1.0.x release instead of an alpha build when considering putting this PR as-is into a release.

Else it looks fine to me :+1:

tobiasKaminsky: Yes, please use stable.

David-Development: Okay, I'll downgrade. I would like to mention that I experienced a couple of issues with version 1.0.0 in the news app.

PR #53, opened 2019-04-06
Migrate to Java 8 / AndroidX / update dependencies

David-Development:
@Unpublished is it okay if we merge your changes back into the main repository? We can also add you as a contributor if you like so you can work on this repo directly as well. 

One question that came up while looking through your changes. The commit called "remove synthetic accessors".. what is the purposes of it? Isn't it better (more appropriate) to have getter methods for them? 

Glad that someone uses it and takes a look at the code! :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,6 +34,10 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
     productFlavors {
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -43,7 +43,9 @@ android {
 }
 
 dependencies {
-    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
+    implementation 'androidx.annotation:annotation:1.0.2'
+    implementation 'androidx.core:core:1.0.1'
+    implementation 'androidx.fragment:fragment:1.0.0'
 
     implementation 'com.google.code.gson:gson:2.8.5'
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,11 @@
 buildscript {
     repositories {
-        jcenter()
         google()
+        jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.3.2'
-        classpath 'com.github.dcendents:android-maven-gradle-plugin:2.0'
+        classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,14 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    lintOptions{
+        // translations are imported from transifex, so no need to check here
+        disable 'MissingTranslation'
+        disable 'ExtraTranslation'
+        disable 'MissingQuantity'
+        disable 'InconsistentArrays'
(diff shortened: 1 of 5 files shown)


Unpublished:
@David-Development 
> @Unpublished is it okay if we merge your changes back into the main repository?

Yes some of them are ready for this while others aren't yet. I'll split them.

> We can also add you as a contributor if you like so you can work on this repo directly as well.

I'd appreciate being added as a contributor.

> One question that came up while looking through your changes. The commit called "remove synthetic accessors".. what is the purposes of it? Isn't it better (more appropriate) to have getter methods for them?

There's more background to this [here](https://academy.realm.io/posts/360andev-jake-wharton-java-hidden-costs-android/) (starting at "Anonymous class (13:06)").
Making affected fields package-private is recommended solution for libraries as well.

Basicly it reduces the method count in the generated dex file. Adding methods would be contrary.
I just measured the impact on this library:
(536 -> 518)
It's small but it can add up since the library is likely to be included in multiple installed apps.

> Glad that someone uses it and takes a look at the code! :)

👍

David-Development:
@Unpublished

> Yes some of them are ready for this while others aren't yet. I'll split them.

Okay, Awesome! :)

> I'd appreciate being added as a contributor.

@tobiasKaminsky @AndyScherzinger can you add @Unpublished as a contributor, please? I don't have the permission to do it.

> There's more background to this here (starting at "Anonymous class (13:06)").
> Making affected fields package-private is recommended solution for libraries as well.
> Basically it reduces the method count in the generated dex file. Adding methods would be contrary.
> I just measured the impact on this library: (536 -> 518)
> It's small but it can add up since the library is likely to be included in multiple installed apps.

Okay, I'll look into this again. After I briefly read an article about synthetic accessors, they talked only about nested classes (which we don't have). I'll take a look at the podcast later it might as well convince me that it's a good idea..
As of now I'm not sure if this an advantage that is worth having (the measured difference is "only" 18). Concerns: External applications can access these variables directly and Codacy complains about code smell.

Unpublished:
> ...only about nested classes (which we don't have)

Yes there are accesses in nested classes, like `Builder` inside `NextcloudRequest` or the anonymous class of `ServiceConnection` inside `NextcloudAPI`.

I understand your concerns and I'm fine with keeping this only in my fork. Though external applications need to access them using the same package. So yes it is easier than using reflection but is still possible to recognize one shouldn't access them directly.

tobiasKaminsky:
> > I'd appreciate being added as a contributor.
> 
> @tobiasKaminsky @AndyScherzinger can you add @Unpublished as a contributor, please? I don't have the permission to do it.

I added you. As a team member of android you should now have access to push branches to this and other android related repos.

Unpublished:
> I added you. As a team member of android you should now have access to push branches to this and other android related repos.

@tobiasKaminsky Thank you for adding me to the android team :)

@David-Development This PR should now be fine to merge.

David-Development:
@Unpublished Awesome, thank you! And thanks for reverting the changes with the modifiers. After watching the video by Jake, I'm still a little torn between whether it's a good or bad thing to get rid of the modifiers..

Unpublished:
@David-Development I think it's okay to just leave these 18 extra methods there :)

Would you merge such a commit on [news-android](https://github.com/nextcloud/news-android) (~ 100 methods there)?

David-Development:
> Would you merge such a commit on news-android (~ 100 methods there)?

@Unpublished I'm not quite sure if I understand that question correctly. Do you have some changes that you want to push?

Unpublished:
@David-Development Sorry, I wasn't clear enough.
I am talking about a [remove synthetic accessors](https://github.com/Unpublished/news-android/commit/7b5e6c9b945cec6de5e2fe8a70c9ca4ec217a9b3) commit for news-android.
Just wanted to know if you feel different about changing the modifiers in an application vs (this) library.

PR #56, opened 2019-04-15
Rename username to userId

tobiasKaminsky:
To clarify and have it in sync across all projects:
userId: internally id, needed for e.g. dav/trashbin/$userId/

loginName: used for authenticate on every login/network connection, determinated by first login (weblogin/old login): can be anything: email, name, name with withespaces

displayName: Name chosen by user, only displayed, no usage at all

There will be soon a matching PR in library/files

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/Constants.java
+++ b/src/main/java/com/nextcloud/android/sso/Constants.java
@@ -22,7 +22,7 @@ package com.nextcloud.android.sso;
 public class Constants {
 
     // Authenticator related constants
-    public static final String SSO_USERNAME = "username";
+    public static final String SSO_USER_ID = "user_id";
     public static final String SSO_TOKEN = "token";
     public static final String SSO_SERVER_URL = "server_url";
     public static final String SSO_SHARED_PREFERENCE = "single-sign-on";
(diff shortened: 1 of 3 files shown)


src/main/java/com/nextcloud/android/sso/Constants.java
```diff
@@ -22,7 +22,7 @@
 public class Constants {
 
     // Authenticator related constants
-    public static final String SSO_USERNAME = "username";
+    public static final String SSO_USER_ID = "user_id";
```
David-Development: Changing this line breaks backward compatibility, right? Is there another option to make the transition soft? And we would have to upgrade third party apps first before you release this into the nextcloud files app.

Maybe add something like this: (in the `AccountImporter.java`)

```java
String userId = future.getString(Constants.SSO_USER_ID);
if (userId == null) {
  // backwards compatibility
  userId = future.getString("username");
}
```

tobiasKaminsky: This is a nice idea.

David-Development:
@tobiasKaminsky is this ready to be merged?

tobiasKaminsky:
Yes, as this ensures that it backward compatible, we can merge this any time
(corresponding PR on files app will also be merged soon)

PR #57, opened 2019-05-02
Add support for Completables

Unpublished:
Otherwise they'll be executed on starting thread (mostly main thread) resulting in crash.


David-Development:
Good point, thank you for the update! :)

Issue #49, opened 2019-03-19
Question: Drop Account Manager


David-Development:
@tobiasKaminsky Just throwing in some thoughts here. As we had a bunch of issues already when trying to figure out which accounts exist, I was thinking about if we can drop the need to use the Android "AccountManager" at all. As of Android 9 getting accounts from another app is even more difficult. Therefore dropping it makes things easier for all involved parties.

**Advantages:**
- reauthentication works without the need to request the account permission again (way easier for client apps as we don't need to restart the login flow)
- Custom dialog for choosing the account (nextcloud design / theme)
- reduce complexity of account handling (in the sso library) 

**Disadvantages:**
- None?

So basically what I'm proposing is:
- Instead of calling the account manager to show us available accounts, we create a new custom activity in the files app for that. [(Reference)](https://github.com/nextcloud/Android-SingleSignOn/blob/master/src/main/java/com/nextcloud/android/sso/AccountImporter.java#L74)
- The files app lists all accounts available (as the files app has the required permission for that)
- The users selects an account here  

This also might be interesting for @desperateCoder @stefan-niedermann

tobiasKaminsky:
Hm. I am not sure if I understand it.
Do you want to get rid of the account handling in files app? And implement our own one?
Or only within SSO?  E.g. https://github.com/nextcloud/Android-SingleSignOn/blob/79df148a8fee888187d75edcaf2ab907de04c8ba/src/main/java/com/nextcloud/android/sso/AccountImporter.java#L92-L104

David-Development:
@tobiasKaminsky The method you mentioned uses the internal Android AccountManager. the call to getAccounts() will only return accounts, that the app currently has permission to access. So if we reinstall the files app, the permission is revoked and therefore the user has to request that permission again. Right now the only purpose of using the Android build in Account Manager is to figure out which accounts are installed and to select one of the accounts. The whole part with requesting / exchanging tokens is handled elsewhere.

That's why I was wondering.. maybe we can throw the Android Account Manager out so we don't have to work around those permission handling stuff? What do you think?

mario:
Personally I find using my own account management/storage compared to Android Account Manager significantly better, even if Android Account Manager has a tighter integration into the OS (which I don't really use or care about).

tobiasKaminsky:
> So if we reinstall the files app, the permission is revoked and therefore the user has to request that permission again.

Even if we use our own account management system, after a reinstallation these infos will be gone, or?

[...]

Issue #46, opened 2019-03-08
Library doesn't recognize the Nextcloud Dev app


strugee:
STR:

1. Install Nextcloud Dev
2. Install Nextcloud News
3 Try to use the SSO option in News

You'll get a message asking you to install the files app, even though the development version is already installed. I'm assuming this is the right place for this issue but I'm happy to refile in the downstream News issue tracker if that's the right place.

David-Development:
Thanks for the feedback. Yes, this is the right place. Indeed, this shouldn't happen.. It shouldn't matter if you have the dev version or the one from google play. 

Couple of questions for you:
- What Android Version are you on?
- Did you compile the nextcloud dev version yourself? 
- What version of the nextcloud news app are you using?

strugee:
I'm on LineageOS 15.1 (Android 8.1). Both apps are installed from F-Droid. News version is 0.9.9.25.

David-Development:
@strugee Thank you :) 

@tobiasKaminsky the problem is, that we don't support the dev version yet. As the package is different from (`com.nextcloud.client`), the sso library won't pick up its existence. 

Is this something we should support? If yes, how do we handle it if both apps are installed and the user configured different accounts in each app? Or the same account? 

Problematic snippets:
https://github.com/nextcloud/Android-SingleSignOn/blob/82bbca0e8ff79eb0cf5690096d84f5c5a7def26a/src/main/java/com/nextcloud/android/sso/AccountImporter.java#L68

https://github.com/nextcloud/Android-SingleSignOn/blob/82bbca0e8ff79eb0cf5690096d84f5c5a7def26a/src/main/java/com/nextcloud/android/sso/helper/VersionCheckHelper.java

tobiasKaminsky:
> Is this something we should support? 

If that is not too much work, why not…

> If yes, how do we handle it if both apps are installed and the user configured different accounts in each app? Or the same account?

When querying & displaying the accounts we then would need to check on both apps and should add (dev) to the dev version accounts.

Accounts from dev and stable app which point to the same user (e.g. tobias@cloud.example.com) should then be distinguished.

strugee:
Closing since #63 got merged and 0.4.0 went out with (according to the GitHub release notes) support for this. If that's not right feel free to reopen! Thanks :thumbsup:

PR #63, opened 2019-05-16
add support for dev app

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/AccountImporter.java
+++ b/src/main/java/com/nextcloud/android/sso/AccountImporter.java
@@ -34,6 +34,10 @@ import android.os.Bundle;
 import android.util.Log;
 import android.widget.Toast;
 
+import androidx.core.app.ActivityCompat;
+import androidx.core.content.ContextCompat;
+import androidx.fragment.app.Fragment;
+
 import com.nextcloud.android.sso.exceptions.AndroidGetAccountsPermissionNotGranted;
 import com.nextcloud.android.sso.exceptions.NextcloudFilesAppAccountNotFoundException;
 import com.nextcloud.android.sso.exceptions.NextcloudFilesAppAccountPermissionNotGrantedException;
@@ -45,11 +49,8 @@ import com.nextcloud.android.sso.ui.UiExceptionManager;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
-
-import androidx.core.app.ActivityCompat;
-import androidx.core.content.ContextCompat;
-import androidx.fragment.app.Fragment;
 
 import io.reactivex.annotations.NonNull;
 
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
SSO has hard dependendcies to Nextcloud File app:
https://github.com/nextcloud/Android-SingleSignOn/blob/2ffb23412081ecf6b1bb81cce92a457f476cac43/src/main/java/com/nextcloud/android/sso/helper/VersionCheckHelper.java#L35-L40

https://github.com/nextcloud/Android-SingleSignOn/blob/53355e83c226679e01acf7192076baf221a806c4/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java#L79-L84

https://github.com/nextcloud/Android-SingleSignOn/blob/c9dc53da8ffd8d7cd264dac8db113507410fd929/src/main/java/com/nextcloud/android/sso/AccountImporter.java#L278-L289

Question is how / if we can also support dev version there.

David-Development:
For reference from our Telegram chat:

> Es gibt wirklich noch einiges zu beachten da der Service entsprechend der jeweiligen App gebunden werden muss. Bin mir noch nicht über Seiteneffekte bei Verwendung von multi Accounts im klaren. Erste Idee ist, dass wir uns für jeden Account speichern müssen von welchem package er ist und dementsprechend den Service binden müssen. Tobi hatte ja bereits die entsprechenden Stellen im Code angegeben. 

> Der Vollständigkeit wegen: hier der Bug-Report der schon seit einigen Monaten offen ist: https://github.com/nextcloud/Android-SingleSignOn/issues/46

@tobiasKaminsky Just thought about #49 .. this will make it quite hard to support both apps (normal and dev) as we need to connect to both apps and get an account list.

tobiasKaminsky:
> need to connect to both apps and get an account list.

as this is a "one time" action, only when adding an account, I think this is ok.

tobiasKaminsky:
This should be ready to test

tobiasKaminsky:
@stefan-niedermann @desperateCoder can you both include this in your app and test?

David-Development:
> as this is a "one time" action, only when adding an account, I think this is ok.

@tobiasKaminsky so for all the requests we'll just use the nextcloud files app? (and not the dev app?)

Does the normal app has access to the account data of the dev app? The actual account data is required for the actual network call to the nextcloud, isn't it?

tobiasKaminsky:
> > as this is a "one time" action, only when adding an account, I think this is ok.
> 
> @tobiasKaminsky so for all the requests we'll just use the nextcloud files app? (and not the dev app?)
> 
> Does the normal app has access to the account data of the dev app? The actual account data is required for the actual network call to the nextcloud, isn't it?

No, I was referring to "fetch accounts from both apps". This is a "one time" action while adding accounts, which does not happen that offen.

As far as I tested, this PR allows to use both apps.
Please test it again so that it is really working as desired. I tested like this:
- official app: account on server A
- dev app: account on server B
- 3rd party app sees both accounts
- 3rd party app can switch between both accounts on the fly and do stuff 
(I tested it with deck)

David-Development:
@tobiasKaminsky I was a little confused at first how the network requests will be made but I just saw that you added some logic to whether to connect to the nextcloud service or the nextcloud dev service. This line made things much clearer to me: https://github.com/nextcloud/Android-SingleSignOn/blob/c67f4a252dcf60feb9e4b607fcf28d12ac3b4333/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java#L79

After running some tests, I couldn't select an account from the dev app as the line `final Account[] accounts = accMgr.getAccountsByType("nextcloud");` in the `AccountImporter` was only returning accounts of the type "nextcloud". After changing it to `final Account[] accounts = accMgr.getAccounts();` the login worked. How did you manage to get in running without that line?




This PR looks good to me! Great work! :) I created a PR with a few minor fixes in #65.

tobiasKaminsky:
> How did you manage to get in running without that line?

Good question, :/
Maybe I tested it wrongly, but your fixes make sense.
Thanks for testing!

desperateCoder:
Yay, thank you! When are you gonna build a release approximately?

David-Development:
@desperateCoder right now! https://github.com/nextcloud/Android-SingleSignOn/releases/tag/0.4.0

desperateCoder:
Thanks buddy! Hope you will be @ the NC conference, so I can offer you a beer for your support! Same for @tobiasKaminsky btw!

desperateCoder:
@David-Development @tobiasKaminsky , just tested it for Deck, works great! Thanks again!

PR #64, opened 2019-05-20
parse query parameters in url and put them into the parameterMap

David-Development:


src/main/java/com/nextcloud/android/sso/api/NextcloudRetrofitServiceMethod.java
```diff
@@ -282,16 +278,32 @@ private Headers parseHeaders(String[] headers) {
...
+            for (String pair : pairs) {
+                int idx = pair.indexOf("=");
+                query_pairs.put(pair.substring(0, idx), pair.substring(idx + 1));
+            }
+
+            // Remove query params from url
+            this.relativeUrl = this.relativeUrl.substring(0, idxQuery);
         }
-        return patterns;
+        return query_pairs;
```
tobiasKaminsky: As we use camel case all the time, I would also use queryPairs

src/main/java/com/nextcloud/android/sso/api/NextcloudRetrofitServiceMethod.java
```diff
@@ -282,16 +278,32 @@ private Headers parseHeaders(String[] headers) {
...
      */
-    /*
-    private static Set<String> parsePathParameters(String path) {
-        Matcher m = PARAM_URL_REGEX.matcher(path);
-        Set<String> patterns = new LinkedHashSet<>();
-        while (m.find()) {
-            patterns.add(m.group(1));
+    private Map<String, String> parsePathParameters() {
+        Map<String, String> query_pairs = new LinkedHashMap<>();
+        int idxQuery = this.relativeUrl.indexOf("?");
```
tobiasKaminsky: According to AS/Annotation relativeUrl can be null and should be checked

tobiasKaminsky:
Some small changes, otherwise :+1: :tada:

PR #65, opened 2019-06-16
Allow dev app fixes

David-Development:


src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
```diff
@@ -76,12 +76,11 @@ public void connect(String type) {
             stop();
         }
 
-        String componentName;
-        if (type != null && type.equalsIgnoreCase(Constants.ACCOUNT_TYPE_DEV)) {
```
David-Development: @tobiasKaminsky The type should never be null and the type is always in lower case, right?

PR #66, opened 2019-07-18
Update README.md to reflect v0.4.0

burningthumb:
I updated the read me to reflect, better, the current version from the maven repository.  It may not be perfect but I think it's at least better than what was there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ private void openAccountChooser() {
 From an Activity
 
 ```java
+@Override
+public void onActivityResult(int requestCode, int resultCode, Intent data) {
+    super.onActivityResult(requestCode, resultCode, data);
             AccountImporter.onActivityResult(requestCode, resultCode, data, this, new AccountImporter.IAccountAccessGranted() {
 
                     NextcloudAPI.ApiConnectedListener callback = new NextcloudAPI.ApiConnectedListener() {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,25 +20,113 @@ repositories {
 }
 
 dependencies {
-    implementation "com.github.nextcloud:Android-SingleSignOn:0.3.0"
+	implementation 'com.google.code.gson:gson:2.8.5'
```
tobiasKaminsky: If this is needed, it should be a dep from SSO

tobiasKaminsky:
Thank you very much for your effort!

@David-Development can you please have a look as you are also using it as "client" and know the code best :-)

README.md
burningthumb: The only thing I know is that the read me has this line of code in onActivityResult:

`NextcloudAPI nextcloudAPI = new NextcloudAPI(context, ssoAccount, new GsonBuilder().create(), callback);`

And if you don't include the implementation 'com.google.code.gson:gson:2.8.5' dependency it does not build.

David-Development: @tobiasKaminsky indeed.. would be better to include it as a dependency of the sso library as it is mandatory.

David-Development: Just opened #67 as a TODO to work on this.

David-Development:
Looks good to me! @burningthumb Thanks for contributing and improving the docs! Hopefully you like this library! If you have any other suggestions on how to improve this library, let us know! :)

PR #71, opened 2019-07-22
I am using SSO from another library so gson need to be imported using api not implementation

burningthumb:
Hopefully I got it right this time, it seems to be working for me now.


burningthumb:
I also updated to the latest versions of the other dependencies so, as far as I can tell, all the dependencies are using the latest versions as of today.

build.gradle
```diff
@@ -64,7 +64,7 @@ dependencies {
     implementation 'commons-io:commons-io:2.6'
 
     implementation 'com.squareup.retrofit2:retrofit:2.5.0'
-    implementation 'com.squareup.okhttp3:okhttp:3.12.3'
+    implementation 'com.squareup.okhttp3:okhttp:4.0.1'
```
Unpublished: @burningthumb Can you revert this commit? It was decided to stay on legacy 3.12.* for now. See discussion [here](https://github.com/nextcloud/Android-SingleSignOn/commit/e781a33e0115ab8c3f4983fed2def1ab924fb9d9#diff-c197962302397baf3a4cc36463dce5ea)

tobiasKaminsky:
> I also updated to the latest versions of the other dependencies so, as far as I can tell, all the dependencies are using the latest versions as of today.

I enabled depend a bot, so that in future we get new PR one for each updated dependency.

burningthumb:
OK.  I reverted the dependency.  I tried to add a comment but got a conflict so I removed it.  I guess you could add a comment to the build.grade file to indicate the reason to not update that dependency in the build.grade itself.

tobiasKaminsky:
I added a comment in #80 
Can you do one last rebase & squash? 

Thank you very much for your help!

burningthumb:
Sure. Done.

David-Development:
@burningthumb Thanks for working on this and sorry for the delay! Also thanks for adding the notes about the `Uri.encode` as I didn't test filenames containing whitespaces before. So this is extremely helpful for other developers trying to integrate this library into their own android apps. Thank you!

PR #87, opened 2019-07-25
check if drone should run: also when build.gradle changed

tobiasKaminsky:


tobiasKaminsky:
Otherwise bumps via dependabot would not be checked

PR #102, opened 2019-09-09
Prevent crash if for some, unknown, reason, mCallback is null

burningthumb:
I don't know why it happened but I got an App crash because mCallback was null.  This was how my code looked when the API is created.
```

private NextcloudAPI.ApiConnectedListener mNextCloudApiCallback = new NextcloudAPI.ApiConnectedListener()
    {
        @Override
        public void onConnected()
        {
        }
        @Override
        public void onError(Exception ex)
        {
        }
    };

```
And this:
`
mNextcloudAPI = new NextcloudAPI(a_context, mSignInAccount, 
                                     new GsonBuilder().create(), mNextCloudApiCallback);
 `

I don't know why it happened but I put a check into AidlNetworkRequest to check for the condition and have not, yet, been able to re-create the crash.  I'm not saying this fixes the problem since I'm still testing it but thought I'd pass it along for you all to think about.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -130,7 +130,7 @@ detekt {
 
 dependencies {
 
-    implementation "androidx.appcompat:appcompat:1.1.0"
+    implementation "androidx.appcompat:appcompat:1.0.2"
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.core:core:1.1.0'
     implementation 'androidx.fragment:fragment:1.1.0'
--- a/build.gradle
+++ b/build.gradle
@@ -130,7 +130,7 @@ detekt {
 
 dependencies {
 
-    implementation "androidx.appcompat:appcompat:1.0.2"
+    implementation "androidx.appcompat:appcompat:1.1.0"
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.core:core:1.1.0'
     implementation 'androidx.fragment:fragment:1.1.0'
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
I think it would be better to make callback @NonNull and check for it during init.
If it is null, we throw directly a IllegalArgumentException so that the other app can handle it properly (retry, show error message to user).

Or do you think that callback got destroyed while usage?

David-Development:
@burningthumb Maybe you can check the following merge request: https://github.com/nextcloud/Android-SingleSignOn/pull/104

Tobias and me fixed some issues where multiple calls to `connect()` caused the app to crash. Please note, that once you call `stop()` on the API, you need to create a new one. You can't re-use the API once you called `stop()` on it.

burningthumb:
As a general programming style I null out objects that should not be re-used.  In the case of Nextcloud, to make sure that I never re-use a stopped reference, whenever I call `stop()` (this happens 2 places in my `Worker` code) on my reference to the API I null it out.  Like this:

```
        if (null != mNextcloudAPI)
        {
            mNextcloudAPI.stop();
            mNextcloudAPI = null;
        }
        mSignInAccount = null;
```

With the nextcloud API, each time I enqueue a new `Worker` I get a new reference that is only used for the lifetime of the `Worker`:

```
    private void setupAPI(Context a_context)
    {
        try
        {
            mSignInAccount = SingleAccountHelper.getCurrentSingleSignOnAccount(a_context);
        }
        catch (Exception e)
        {
            mSignInAccount = null;
        }

        if (null != mSignInAccount)
        {
            mNextcloudAPI = new NextcloudAPI(a_context, mSignInAccount, new GsonBuilder().create(), mNextCloudApiCallback);
        }
    }
```

So this, I think, is not the same issue but I will look at #104 to be sure.  Thank you.

David-Development:
Does this issue still exist? Or did our changes fix the crashes you were experiencing?

burningthumb:
I have not had time to retest it.

David-Development:
Closing this for now. Please feel free to reopen if it happens again!

PR #104, opened 2019-09-13
fix app crash when connect is called multiple times

David-Development:
ref: https://github.com/nextcloud/Android-SingleSignOn/pull/102


tobiasKaminsky:
@burningthumb can you try this with your app to see if this fixes your crash?

Issue #106, opened 2019-09-20
Make import list filterable


stefan-niedermann:
When the list of importable accounts is displayed, it would be great to be able to filter the ones out that are actually imported already.

workflow in 3rd party app like deck or news basically would be:

display dialog with filter = []

-> displays user A and user B

import user A

display dialog with filter = [A]

-> displays only user B

import user B;

cc @nextcloud/designers

PR #108, opened 2019-09-24
Add support to exclude existing accounts

tobiasKaminsky:
Fix https://github.com/nextcloud/Android-SingleSignOn/issues/106

TODO
- [ ] check other usages of pickNewAccount

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/AccountImporter.java
+++ b/src/main/java/com/nextcloud/android/sso/AccountImporter.java
@@ -34,6 +34,7 @@ import android.os.Bundle;
 import android.util.Log;
 import android.widget.Toast;
 
+import androidx.annotation.Nullable;
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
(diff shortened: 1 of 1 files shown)


David-Development:
I didn't have time yet to test the proposed changed. One question that came to my mind: Does this PR work when both apps have different signatures? I thought that `findAccounts` won't return the existing accounts if the user hasn't explicitly granted access to the accounts previously..?

stefan-niedermann:
> I thought that findAccounts won't return the existing accounts if the user hasn't explicitly granted access to the accounts previously..?

This seems to be correct. I just tried it in a branch of the Notes app (given i have multiple accounts, and one of them already using in Notes), where `findAccounts()` will always return only the one account i already have approved.

Well, if `alloweableAccounts` match the result of `findAccounts`, the result will be, that the SSO lib assumes, there is no other account for import available and always opens the Weblogin-flow to add a completely new account.

tobiasKaminsky:
findAccounts() should find all Accounts, and then all existingAccounts should be removed from it.
Hm. Maybe this is again about wrong testing as I tested it with the same Android Studio and not with different signatures.

If findAccounts cannot find all Accounts by Files app (and if we do not find a way), this is useless.

tobiasKaminsky:
So, as far as I understand this does not work if we have different certificates, or?
Then let us postpone this until we have our own account manager in files app.

@David-Development ?

PR #128, opened 2019-10-24
SSO: return response headers

tobiasKaminsky:
Upon release (0.5.0) I will add a deprecation warning/info.

Then 3(?) months later a new version (0.6.0) will remove the old system.

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
+++ b/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
@@ -11,8 +11,8 @@ import android.os.ParcelFileDescriptor;
 import android.os.RemoteException;
 import android.util.Log;
 
-import androidx.annotation.NonNull;
-
+import com.google.gson.Gson;
+import com.google.gson.internal.LinkedTreeMap;
 import com.nextcloud.android.sso.Constants;
 import com.nextcloud.android.sso.aidl.IInputStreamService;
 import com.nextcloud.android.sso.aidl.IThreadListener;
@@ -27,7 +27,11 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
+
+import androidx.annotation.NonNull;
 
 import static com.nextcloud.android.sso.exceptions.SSOException.parseNextcloudCustomException;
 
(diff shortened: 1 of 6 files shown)


stefan-niedermann:
Is this implemented as a breaking change, isn't it? Maybe we should add some documentation and ping all known users?

tobiasKaminsky:
I will, once you tested it, try to do it non breaking change.

tobiasKaminsky:
If not, it will be in release notes and a higher version bump

src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
```diff
@@ -218,10 +247,128 @@ public void onThreadFinished(Thread thread) {
         return output;
     }
 
+    /**
+     * DO NOT CALL THIS METHOD DIRECTLY - use @link(performNetworkRequest) instead
```
David-Development: Link should be adjusted `performNetworkRequest` --> `performNewNetworkRequest`

src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
```diff
@@ -218,10 +247,128 @@ public void onThreadFinished(Thread thread) {
...
+     *
+     * @param request
+     * @return
+     * @throws IOException
+     */
+    private ParcelFileDescriptor performNewAidlNetworkRequest(NextcloudRequest request,
```
David-Development: Maybe we should use some kind of versioning instead of `New`. E.g. `performAidlNetworkRequestV2`

src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
```diff
@@ -115,10 +115,39 @@ public void stop() {
      * @return InputStream answer from server as InputStream
      * @throws Exception or SSOException
      */
-    public InputStream performNetworkRequest(NextcloudRequest request) throws Exception {
+     public InputStream performNetworkRequest(NextcloudRequest request) throws Exception {
         return networkRequest.performNetworkRequest(request, null);
     }
 
+    public Response performNewRequest(final @NonNull Type type, NextcloudRequest request) throws Exception {
```
David-Development: same here.. maybe `performRequestV2`?

src/main/java/com/nextcloud/android/sso/api/Response.java
```diff
@@ -0,0 +1,35 @@
+package com.nextcloud.android.sso.api;
+
+import java.io.InputStream;
+import java.util.ArrayList;
+
+import androidx.annotation.Nullable;
+
+public class Response {
+    private InputStream inputStream;
```
David-Development: maybe we should rename `inputStream` to `body`? (Might be more obvious for using it. In the Nextcloud Android App this is called "response body stream" as well..) https://github.com/nextcloud/android/blob/8955709705a8ae937403b1102a1d11cccc6b9258/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java#L298)

src/main/java/com/nextcloud/android/sso/helper/Okhttp3Helper.java
```diff
@@ -44,4 +44,13 @@ public static ResponseBody getResponseBodyFromRequest(NextcloudAPI nextcloudAPI,
         return ResponseBody.create(null, "");
     }
 
+    public static ResponseBody getResponseBodyFromNewRequest(NextcloudAPI nextcloudAPI, NextcloudRequest request) {
```
David-Development: should probably be renamed to `getResponseBodyFromRequestV2` as well?

src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
```diff
@@ -218,10 +247,128 @@ public void onThreadFinished(Thread thread) {
...
     }
+
+    private ExceptionResponse deserializeNewObject(InputStream is) throws IOException, ClassNotFoundException {
+        ObjectInputStream ois = new ObjectInputStream(is);
+        ArrayList<PlainHeader> headerList = new ArrayList<>();
+        Exception exception = (Exception) ois.readObject();
+
+        if (exception == null) {
+            String headers = (String) ois.readObject();
+            ArrayList list = new Gson().fromJson(headers, ArrayList.class);
```
David-Development: In terms of performance it might make sense to create one instance of the Gson object and reuse it in all requests? If I remember correctly creating a new Gson parser is a quite costly operation. And I think we can even parse those values already as string by using:

```java
list = gson.fromJson(headers, new TypeToken<List<String>>(){}.getType());
```

https://stackoverflow.com/a/12384156/11405479

src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
tobiasKaminsky: V2 is a good idea, I will change it everywhere.

src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
tobiasKaminsky: Those are name value pairs, not plain strings, so this is not possible

Issue #124, opened 2019-10-11
Remove select_account_unknown_error_toast toast


stefan-niedermann:
In `AccountImporter::onActivityResult` you are making a toast with the (translated) text `select_account_unknown_error_toast`. This feels very strange, **please remove it.**

This can happen e.g. if the user presses the back button while he is picking an account. In this case, it is disproportionate to display a text that might make the user feel unsecure as it implies an error or an mistake - but it does not have to be necessarily. **Just print a log** or you might even throw an exception, but it is up to the 3rd-party-app to handle an abortion of an import. 

![ezgif-5-b3c0e91c0bf0](https://user-images.githubusercontent.com/4741199/66633800-f1616c80-ec0b-11e9-9580-2f9ba6d6af3d.gif)

tobiasKaminsky:
Totally true, we should return an exception to 3rd party app.

PR #129, opened 2019-10-25
throw exception if SSO import is cancelled

tobiasKaminsky:
Fix https://github.com/nextcloud/Android-SingleSignOn/issues/124

@David-Development this is a breaking change (the 3rd party user have to catch the exception).

I will add this to release notes.


David-Development:
Looks good to me! Already created a ticket in the nextcloud-news repo so that I don't forget to add this

PR #163, opened 2020-01-21
add more debugging output / destroy api when reconnect fails

### LINKED ISSUE #162: NextcloudApiNotRespondingException
By now two apps are having issues with the `NextcloudApiNotRespondingException` Bug.

- https://github.com/stefan-niedermann/nextcloud-notes/issues/653
- https://github.com/nextcloud/news-android/issues/769

A few days ago I was experiencing the same issue on my device (I think right after I updated the news app via the play store). Several restarts of the app did not help. Only setting up the account again solved the problem for me. Still not sure if the Nextcloud Files App has something to do with it or if the problem is on the "client" side. @stefan-niedermann Do you have new insights into why this error happens? Or any ideas if the files app is involved in this? 

Maybe some old Service Connections are reused which are broken? But that wouldn't make sense in terms of the app restarts I performed. Maybe restarting the phone helps..? That would be interesting to find out.

David-Development:
Closes #162

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
+++ b/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
@@ -51,7 +51,7 @@ public class AidlNetworkRequest extends NetworkRequest {
      */
     private ServiceConnection mConnection = new ServiceConnection() {
         public void onServiceConnected(ComponentName className, IBinder service) {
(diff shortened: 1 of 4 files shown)


David-Development:
@stefan-niedermann @desperateCoder did you guys manage to test the changes proposed in the MR?

stefan-niedermann:
Not yet, sry

PR #171, opened 2020-02-10
Added more sample apps

stefan-niedermann:
PS: @David-Development the sample links of the News app are outdated.


David-Development:
thank you for the hint @stefan-niedermann. just updated the links

PR #225, opened 2020-07-24
Update tests for V2

Unpublished:
Added lenient to suppress unused MockitoHints (which seem to be wrong) as at least the second stub is needed to pass tests again.
```
[MockitoHint] 1. Unused -> at com.nextcloud.android.sso.api.TestRetrofitAPI.setUp(TestRetrofitAPI.java:61)
[MockitoHint] 2. Unused -> at com.nextcloud.android.sso.api.TestRetrofitAPI.setUp(TestRetrofitAPI.java:62)
```

@tobiasKaminsky You maybe want to continue on #77 as drone still doesn't compile or execute tests.

PR #226, opened 2020-07-24
Run basic test via drone

tobiasKaminsky:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,4 +1,46 @@
 kind: pipeline
+type: docker
+name: tests
+
+steps:
+  - name: gplay
+    image: nextcloudci/android:android-49
+    environment:
+      LOG_USERNAME:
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -1,4 +1,47 @@
 kind: pipeline
+type: docker
+name: tests
+
+steps:
+  - name: gplay
+    image: nextcloudci/android:android-49
+    privileged: true
```
Unpublished, line 8: I think you can remove this line AFAIR it is only needed for the emulator

Unpublished: Confirmed on `testDroneFix` branch, I picked this PR without the priviliged line into mine to see if tests actually pass (they passed 🎉 ).

tobiasKaminsky:
Merged via #225

PR #235, opened 2020-08-08
fix file uploads vor networkRequests V2

desperateCoder:
Signed-off-by: @desperateCoder

This fixes the File Upload

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
+++ b/src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
@@ -133,11 +133,14 @@ public class NextcloudAPI {
              Reader targetReader = new InputStreamReader(os)) {
             if (targetEntity != Void.class) {
                 result = gson.fromJson(targetReader, targetEntity);
-                /*
-                if (result != null) {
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
We are in production with this branch of the SSO since a week and it seems to work flawlessly. I vote for merging and releasing this until we [have a better solution](https://github.com/nextcloud/Android-SingleSignOn/pull/236).

desperateCoder:
@David-Development see the last commit (https://github.com/nextcloud/Android-SingleSignOn/commit/7d0b95ebf1ba979870abca02cea16d0c7187d772)

It doesn't solve the `NOTHING` "problem", but I don't consider it as one to be honest. If you want to chat about this solution, I am willing to tell you in detail, why I think this is a viable solution. But we had some blind spots with this, I tried to fix with the commit mentioned above, and I think, we can't do better here. Whats your opinion on that?

CC: @stefan-niedermann

David-Development:
Ahah oh man.. 😅 In which cases does the gson parsing fail? 

You convinced me that we need the `NOTHING` return value so all in all it looks good to me 👍

tobiasKaminsky (approved):

PR #236, opened 2020-08-08
fix issues with Void return type due to rxjs map operator null checks

David-Development:


desperateCoder:
This one will still die (in most cases my code comes across this line...):

https://github.com/nextcloud/Android-SingleSignOn/blob/95ced6841da33239e699c0033f88582b5bcb15eb/src/main/java/com/nextcloud/android/sso/api/NextcloudRetrofitServiceMethod.java#L190

src/test/java/com/nextcloud/android/sso/api/TestRetrofitAPI.java
```diff
@@ -64,6 +64,16 @@ public void setUp() {
...
     }
 
+    @Test
+    public void callWithNoReturnType() throws Exception {
+        mApi.callWithNoReturnType();
+        NextcloudRequest request = new NextcloudRequest.Builder()
+                .setMethod("GET")
+                .setUrl(mApiEndpoint + "callWithNoReturnType")
+                .build();
+        verify(nextcloudApiMock).performRequestV2(eq(retrofit2.Call.class), eq(request));
```
desperateCoder, line 74: Doesn't this just verify the call is made? The problem is returning the async result, so I'm not sure if this a viable test... Does this test break if you bring the `map()` back?

David-Development: Well.. If you take a look at the map operator, it has a null-check inside (`Object.ensureNotNull()`). Therefore we can't use it 🤔

David-Development: Yes, that is indeed a problem. The data is coming from a mock and therefore the observable always returns an empty observable. If we want to add more checks I think we'd have to add dummy objects for all types.. right?

desperateCoder: I think a mocked test isn't the way to go here, at least we are mocking the wrong objects here... The mock would need to be much deeper, somewhere at the NC Files app interface I guess. We need a mock which instantly triggers the callbacks to see, if anything dies in this chain, but I didn't dig in this far yet. I'll have a look!

src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
```diff
@@ -119,9 +118,12 @@ public void stop() {
         try (Reader targetReader = new InputStreamReader(responseStream)) {
             if (targetEntity != Void.class) {
                 return gson.fromJson(targetReader, targetEntity);
+            } else {
```
desperateCoder, line 121: This is a safe `ClassCastException`, since we are casting the result... Correct me if I'm wrong, but i think you cast the `Object` somewhere to `Void`, because I definitely don't. I used to get, what I defined I want to get, so the casting is done SSO-Internally, see comment below...

src/main/java/com/nextcloud/android/sso/api/NextcloudRetrofitServiceMethod.java
```diff
@@ -188,14 +188,13 @@ public T invoke(NextcloudAPI nextcloudAPI, Object[] args) throws Exception {
...
                 return (T) nextcloudAPI.performRequestObservableV2(typeArgument, request).map(r -> r.getResponse());
-
             } else if(ownerType == Call.class) {
                 Type typeArgument = type.getActualTypeArguments()[0];
                 return (T) Retrofit2Helper.WrapInCall(nextcloudAPI, request, typeArgument);
             }
         } else if(this.returnType == Observable.class) {
             // Observables without any type information (see above for Observables with type info)
-            return (T) nextcloudAPI.performRequestObservableV2(Void.class, request); //.map(r -> r.getResponse());
+            return (T) nextcloudAPI.performRequestObservableV2(Void.class, request).map(r -> r.getResponse());
```
desperateCoder, line 197: here we go, here is the  `ClassCastException` going to happen

src/test/java/com/nextcloud/android/sso/api/TestRetrofitAPI.java
David-Development: I agree, the tests here are mostly to check that the type and url was extracted correctly and that the interface was called. However return types are not considered. Thank you for looking into it!

src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
David-Development: Ugh.. me being naive while thinking that I could just throw in some typescript-style code 🤷‍♂️ 😅  I think you might be right about the casting thing. However it would be interesting to see if it actually fails 🙈

David-Development: Hm yes, I get that point. I think I'm also good with the nothing option then. However I'm just thinking about other case. So if you specify `List` without a type, it'll be interpreted as `Void` as well, right? But what if I'm too lazy as a programmer to write the type but I still want to somehow receive my response object..? Maybe I'm overthinking it a little here..

desperateCoder: > So if you specify List without a type, it'll be interpreted as Void as well, right? But what if I'm too lazy as a programmer to write the type but I still want to somehow receive my response object..?

Nope, would be a `List<Object>` you could cast, if your're brave enough to guess the actual type. Would work in this case. The problem with your solution is only the void type, which is broken here. So we're back at the beginning of our problem :rofl:. I'll have another look at it, maybe this weekend and see what I can do about this.

David-Development:
@desperateCoder Did you have a chance to fight the unit tests yet? 🙈 Or do you prefer that we close this pr here and create a new issue instead?

desperateCoder:
@David-Development Actually not, since @stefan-niedermann keeps throwing feature requests at my head... If you say you really need my help, I'll see what I can do, although I hate testing 😅

AndyScherzinger:
@David-Development @desperateCoder @stefan-niedermann I guess this PR is obsolete given the lately merged new approach brought in by Stefan via #542

stefan-niedermann:
obsolete from my PoV

PR #270, opened 2020-10-21
Use SSO auth token to let files app detect if new account was requested via SSO

tobiasKaminsky:
Ref: https://github.com/nextcloud/android/pull/7124


David-Development:
@tobiasKaminsky It has unresolved conflicts. Other than that, it looks good to me!

PR #282, opened 2020-11-16
Add queries to Manifest so that lib can find Files App for SDK>=30

### LINKED ISSUE #277: Package visibility issue when targetSdk="30"
Android [has changed the package visibility](https://developer.android.com/about/versions/11/privacy/package-visibility) starting with Android 11.

This means basically, if a 3rd party app increases its `targetSdk` attribute in the `AndroidManifest.xml`, then it will no longer be able to find the Nextcloud Files app. `AccountImporter#appInstalledOrNot` will always return `false`, because `PackageManager#getPackageInfo` will not return any infos about 3rd party packages.

I am currently trying to find a solution based on the [`queries` mentioned in step `7`](https://developer.android.com/about/versions/11/privacy/package-visibility#test) and will keep you up to date. Though this should not be done by the 3rd party app, but in the [`AndroidManifext.xml` of the SSO library](https://github.com/nextcloud/Android-SingleSignOn/blob/master/src/main/AndroidManifest.xml).

tobiasKaminsky:
Fix #277 

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -2,6 +2,11 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.nextcloud.android.sso">
 
-    <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.GET_ACCOUNTS" />
+
+    <queries>
+        <package android:name="com.nextcloud.client" />
+        <package android:name="com.nextcloud.android.beta" />
+    </queries>
 </manifest>


stefan-niedermann:
Nice, i will try to test & verify this solution!

stefan-niedermann (approved):
Seems to work (tested with the Deck app).

However i would feel safer if someone (@matiasdelellis or @David-Development ?) could verify this :slightly_smiling_face:

matiasdelellis:
Hi @stefan-niedermann 
I have a user who reported it too.  I gladly try it..  However, is there some tagged version or I have to incorporate the library into the code to test it?

stefan-niedermann:
@matiasdelellis you can simply use this as dependency:
```
implementation 'com.github.nextcloud:Android-SingleSignOn:query-SNAPSHOT'
```
instead of the default SSO dependency.

Ideally you could build the APK with it and provide it to the user who reported it to you, so we can test it with a device which actually **has** this issue :slightly_smiling_face:  (I am only able to test it on emulated devices.... all very theoretically...)

matiasdelellis:
Done on https://github.com/matiasdelellis/nextcloud-quicknotes-android/issues/4
I await user feedback.

Thanks. 😄

stefan-niedermann:
@matiasdelellis just to be sure - have you also raised the `compileSdk` and `targetSdk` to `30` in your test build apk again? :laughing:

matiasdelellis:
D'Oh!. 😓 
I did it now. 🙈

matiasdelellis:
Hi,
I did two PR and I have the report of two users.

https://github.com/matiasdelellis/nextcloud-quicknotes-android/pull/9
* SingleSignOn:query-SNAPSHOT
* compileSdkVersion 29
* minSdkVersion 21
* Dont set targetSdkVersion. So implicitly at 21 by the above setting.

Both users were able to find the nextcloud client, select the user and use the App.

https://github.com/matiasdelellis/nextcloud-quicknotes-android/pull/10
* SingleSignOn:query-SNAPSHOT
* compileSdkVersion 30
* minSdkVersion 21
* targetSdkVersion 30

Both users reported the same error. :disappointed: 

However, they reported that the client in master also works without the `SingleSignOn:query-SNAPSHOT`. I will try to inquire into this. :confused:

tobiasKaminsky:
Any update on this?

Issue #280, opened 2020-11-09
A disabled application results in an unmanageable exception


matiasdelellis:
The Nextcloud server when you access a disabled application, it redirects you to the main page. For example, with the tutorial API.

```
[matias@nube nextcloud]$ curl -v -u user:password -X GET -H "OCS-APIRequest: true" -H "Accept: application/json" https://delellis.com.ar/index.php/apps/notestutorial/api/0.1
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying 127.0.0.1:443...
* Connected to delellis.com.ar (127.0.0.1) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use http/1.1
* Server certificate:
*  subject: CN=delellis.com.ar
*  start date: Aug 20 21:44:24 2020 GMT
*  expire date: Nov 18 21:44:24 2020 GMT
*  subjectAltName: host "delellis.com.ar" matched cert's "delellis.com.ar"
*  issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
*  SSL certificate verify ok.
* Server auth using Basic with user 'matias'
> GET /index.php/apps/notestutorial/api/0.1 HTTP/1.1
> Host: delellis.com.ar
> Authorization: Basic bWF0aWFzOnBlbmNpb25hMg==
> User-Agent: curl/7.71.1
> OCS-APIRequest: true
> Accept: application/json
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
[...]

PR #288, opened 2020-12-04
Fix retrofit call error

David-Development:
Fix https://github.com/nextcloud/Android-SingleSignOn/issues/280

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/helper/Retrofit2Helper.java
+++ b/src/main/java/com/nextcloud/android/sso/helper/Retrofit2Helper.java
@@ -9,12 +9,14 @@ import com.nextcloud.android.sso.exceptions.NextcloudHttpRequestFailedException;
 import java.lang.reflect.Type;
 
 import okhttp3.MediaType;
+import okhttp3.Protocol;
 import okhttp3.Request;
 import okhttp3.ResponseBody;
 import okio.BufferedSource;
 import retrofit2.Call;
 import retrofit2.Callback;
 import retrofit2.Response;
+
 
 /**
  *  Nextcloud SingleSignOn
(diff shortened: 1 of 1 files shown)


David-Development:
@tobiasKaminsky I'm actually not 100% sure whether we should call `onFailure` or return the error response as it is right now. 
I think returning error response is probably better as it is not really a failure when an app is not enabled  - and therefore we are returning an "error" response indicating that the app is not available. 

TLDR; I think it is good as it is right now.

src/main/java/com/nextcloud/android/sso/helper/Retrofit2Helper.java
```diff
@@ -99,6 +101,19 @@ public boolean isCanceled() {
...
+            private Response<T> convertExceptionToResponse(int statusCode, String errorMessage) {
+                ResponseBody body = ResponseBody.create(null, errorMessage);
+                return Response.error(
+                        body,
+                        new okhttp3.Response.Builder()
+                                .body(body)
+                                .code(statusCode)
+                                .message(errorMessage)
+                                .protocol(Protocol.HTTP_1_1)
+                                .request(new Request.Builder().url("http://localhost/" + nextcloudRequest.getUrl()).build())
```
tobiasKaminsky, line 114: Is this hardcoded localhost correct?

David-Development: It's never used anywhere (it's just needed for the error message). I copied this from the reotrofit lib and they are using localhost as well (hardcoded).

tobiasKaminsky (approved):

tobiasKaminsky:
@matiasdelellis  can you give this a test run?

matiasdelellis:
HI both,
Tomorrow test that.. Thanks.. 😄

matiasdelellis:
Hi both,
Great.. Now get a valid empty response, and I can evaluate the redirect code to know what is happening.

Thanks you so much.. 😄

PR #300, opened 2021-01-19
Fix leaking parcel file descriptors

Unpublished:
Caused `A resource failed to call close` warnings to be logged.

Found via [StrictMode](https://developer.android.com/reference/android/os/StrictMode)

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
+++ b/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
@@ -11,13 +11,10 @@ import android.os.ParcelFileDescriptor;
 import android.os.RemoteException;
 import android.util.Log;
 
-import androidx.annotation.NonNull;
-
 import com.google.gson.Gson;
 import com.google.gson.internal.LinkedTreeMap;
 import com.nextcloud.android.sso.Constants;
 import com.nextcloud.android.sso.aidl.IInputStreamService;
-import com.nextcloud.android.sso.aidl.IThreadListener;
 import com.nextcloud.android.sso.aidl.NextcloudRequest;
 import com.nextcloud.android.sso.aidl.ParcelFileDescriptorUtil;
 import com.nextcloud.android.sso.exceptions.NextcloudApiNotRespondingException;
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Thank you for the contribution! Really looking forward to it [since we also experiencing this issue](https://github.com/stefan-niedermann/nextcloud-deck/issues/539) :partying_face:

David-Development:
Great work, thank you! 🙏

David-Development (approved):

Unpublished (commented):
Hold on, I probably found another one

src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
```diff
@@ -164,16 +163,21 @@ private void waitForApi() throws NextcloudApiNotRespondingException {
...
+                if (response.getException().getMessage() != null) {
+                    throw parseNextcloudCustomException(response.getException());
+                }
+                throw response.getException();
             }
-            throw response.getException();
+            return new Response(os, response.headers);
+        } catch (Exception e) {
+            os.close();
+            throw e;
```
Unpublished, line 179: Don't like this...I'm open to suggestions

tobiasKaminsky: Remove catch and put os.close() in finally?
Then Exception will be thrown, but also os.close() is called?

(not tested, but should work)

Unpublished: That was my initial solution. Then I noticed `os` is returned and is intended to stay open, it only needs to be closed on an exception when it isn't returned anymore.

tobiasKaminsky: Ah. Then I think the solution is good.
Can you add an info that os normally should stay open, but only be closed when handling an Exception?

Unpublished: Something like [this](https://github.com/nextcloud/Android-SingleSignOn/compare/e1b80f18df5b453ea2367937fe004ba0ebfbae87..4b1a3cb821f61b3ac604b54a102efcf5ef04a6ca)?

tobiasKaminsky:
> Hold on, I probably found another one

Is this still valid?

Unpublished (commented):
Nope, this is ready to be merged :)

PR #315, opened 2021-03-09
l10n: Delete dot

Valdnet:


tobiasKaminsky (approved):

Valdnet:
@rakekniven I was also wondering if there should be a period at the end of the sentence, even though it is an error message.
PR: #317

PR #376, opened 2021-08-22
Fix #353 - add instructions on how to disable battery optimizations / increase compileSDK

### LINKED ISSUE #353: Suggest disabling battery optimizations if Nextcloud API is not responding
This library contains an error message that can be displayed when the Nextcloud API is not responding asking the user to report that issue. We already figured out, that this is due some battery optimizations.. therefore we should improve that message (and probably also the dialog design) to include information about how to disable battery optimizations.

I'm talking about this line: 
https://github.com/nextcloud/Android-SingleSignOn/blob/master/src/main/res/values/strings.xml#L38

David-Development:
In case the Nextcloud Files API is not responding - this dialog will be showed. It'll take the user straight to the settings screen where they can disable the battery optimizations. 

I also raised the compile SDK version to 30.


src/main/res/values/strings.xml
```diff
@@ -35,7 +35,9 @@
     <string name="nextcloud_files_app_not_supported_message" formatted="true">Your Nextcloud Files app currently does not support the single sign on feature. Please update it: %1$s</string>
 
     <string name="nextcloud_files_api_not_responding_title">Error</string>
-    <string name="nextcloud_files_api_not_responding_message">Nextcloud Files app API is not responding. Please report this issue.</string>
+    <string name="nextcloud_files_api_not_responding_message">Nextcloud Files app API is not responding. This is most likely due to battery optimizations on your phone. Try to disable the battery optimization for both apps (Nextcloud Android and this app). In all other cases please try to clear the storage of both apps.</string>
+
+    <string name="nextcloud_files_api_not_responding_open_battery_optimization_settings">Battery Opt. Settings.</string>
```
stefan-niedermann, line 40: Remove trailing dot

stefan-niedermann (changes requested):

src/main/res/values/strings.xml
David-Development: good catch, thank you! 🙇

build.gradle
```diff
@@ -61,8 +61,8 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
```
stefan-niedermann, line 65: Please also adjust [the README.md](https://github.com/nextcloud/Android-SingleSignOn#1-add-this-library-to-your-project) which still hints to Java 8.

stefan-niedermann (changes requested):
Sorry buddy, one more thing 😉

build.gradle
David-Development: @stefan-niedermann If I got it right you only need it if you are using gradle in Version 7.0.0 or higher, right? And it's not backwards compatible with Gradle 4.2, right?

stefan-niedermann (approved):

tobiasKaminsky (approved):

PR #379, opened 2021-08-30
Add troubleshooting hint

stefan-niedermann:
for quickly appearing and disappearing menus when attempting to select an account based on multiple experiences with Play Store users.


David-Development (approved):
Does this actually happen? What about using the min-version feature when initializing SSO? It should show an error in those cases.

stefan-niedermann:
> Does this actually happen?

Here two recent samples of ratings we got in the deck app. Note the identical description, explaining that the account chooser prompt appears and immediately disappears.

grafik
grafik

In both cases upgrading the files app solved the issue:

grafik
grafik

> What about using the min-version feature when initializing SSO?

You are absolutely right, this will be the next steps for us: Trying various versions of the files app and if we found the version that is incompatible, we can add a min-version check.

Though "The account chooser disappears quickly" was a quite confusing problem description for us and we did not know in the first place where to start. If upgrading the files app to the most recent version solves or works around this issue, it seems to be a nice addition to the troubleshooting section for me. And it naturally leads to implementing a min-version check, but in the first place it helps you to *know* that you *have to*  implement a min-version check :smile:

PR #407, opened 2021-11-06
#113 Remove deprecated v1 system

stefan-niedermann:
This removes all deprecated APIs from the SSO lib regarding the V1 system.

Hints for review:

- The old methods need to stay in `IInputStreamService.aidl` though, otherwise the sample app breaks. Ideas for a smooth migration are welcome, but i doubt that it is worth the efforts.
- `NextcloudRetrofitServiceMethod#invoke` always used `Okhttp3Helper.getResponseBodyFromRequest`, i moved this to `Okhttp3Helper.getResponseBodyFromRequestV2`
- `AidlNetworkRequest#performAidlNetworkRequest` used the following code part. The V2 pendant never had this. Shall we move this to the V2 or just ignore? (I also marked it as `TODO` in the code…)
```java
if (mDestroyed) {
    throw new IllegalStateException("Nextcloud API already destroyed. Please report this issue.");
}
```

- [ ] Unit tests
- [ ] Review

### DIFF
--- a/lib/src/main/aidl/com/nextcloud/android/sso/aidl/IInputStreamService.aidl
+++ b/lib/src/main/aidl/com/nextcloud/android/sso/aidl/IInputStreamService.aidl
@@ -18,11 +18,7 @@
 package com.nextcloud.android.sso.aidl;
 
 interface IInputStreamService {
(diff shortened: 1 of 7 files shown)


stefan-niedermann:
Phew, seems to be a bit more complicated than i thought... I'll close this again, if someone wants to pick up this again, i'll keep the branch here for a few days before deleting it...

Issue #409, opened 2021-11-15
Support .qa package id of the files app


stefan-niedermann:
Currently only prod and dev are supported, we should extend the list by the `.qa` package id of the files app, so one can test combinations with CI/CD builds of the files app.

PR #410, opened 2021-11-15
#409 Support .qa package id of the files app

stefan-niedermann:
- Move loose Package ID and Account Type from `Constants` class to own `Enum` (`FilesAppType`)
- Mark old constants as `@Deprecated`, but keep them to maintain backward compatibility
- Clean up some methods which internally always used `prod` as target, even though we already supported `dev`. They will now be callable with an explicit `FilesAppType` (also full backward compatible)

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -41,6 +41,7 @@ android {
         }
     }
     compileOptions {
+        coreLibraryDesugaringEnabled true
         sourceCompatibility JavaVersion.VERSION_11
         targetCompatibility JavaVersion.VERSION_11
     }
@@ -97,6 +98,8 @@ detekt {
 }
 
 dependencies {
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
+
     implementation "androidx.appcompat:appcompat:1.3.1"
     implementation 'androidx.annotation:annotation:1.3.0'
     implementation 'androidx.core:core:1.7.0'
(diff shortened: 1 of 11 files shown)


lib/src/main/java/com/nextcloud/android/sso/model/FilesAppType.java
```diff
@@ -0,0 +1,33 @@
+package com.nextcloud.android.sso.model;
```
tobiasKaminsky, line 1: Please add licence

lib/build.gradle
```diff
@@ -41,6 +41,7 @@ android {
         }
     }
     compileOptions {
+        coreLibraryDesugaringEnabled true
```
tobiasKaminsky, line 44: What's this about?

lib/src/main/java/com/nextcloud/android/sso/model/FilesAppType.java
stefan-niedermann: Should i also add it to other files like
- `AidlNetworkRequest`
- `VersionCheckHelper`
- `UiExceptionManager`
- and various other existing files where it is not present?

stefan-niedermann: I added the license header to the files i saw without, a few of them have them on a different place, that's why i missed it....

lib/build.gradle
stefan-niedermann: Concrete usage in this PR: https://github.com/nextcloud/Android-SingleSignOn/pull/410/files#diff-61442b83bbc4d761051337c72c535ce0e5a14da8c3aa3c5767f76bef95b30652R78

```java
private static final String[] ACCOUNT_TYPES = Arrays.stream(FilesAppType.values()).map(a -> a.accountType).toArray(String[]::new);
```

tobiasKaminsky: Totally fine, I was not aware of this.

tobiasKaminsky (approved):

Issue #133, opened 2019-10-28
Remove old v1 system


tobiasKaminsky:
- remove old v1 system to only support one system (with Headers)

PR #411, opened 2021-11-16
#133 Remove deprecated methods

stefan-niedermann:
Another attempt to remove all deprecated stuff (mainly the `v1` API - #133).

The sample app works pretty well, but i recommend to put some weight into the review of the `Okhttp3Helper` - it seems like it *always* used `v1` in the past and never `v2`. I changed that.

This is a **Breaking Change** ~~and will cause the SSO library to only work with the files app that supports `v2`. We should check which version of the files app introduced support for `v2` for the release notes.~~ for 3rd party app developers ,but since we keep filling the `parameter` property along to `parameterV2` in `NextcloudRequest`, it might still even work with older files app versions (?).


David-Development (approved):

David-Development:
Thank you Stefan! 🙇‍♂️ 🥳

Issue #400, opened 2021-11-04
Make NextcloudAPI constructor less verbose


stefan-niedermann:
# Issue

When instantiating the `NextcloudAPI` we currently have some quite verbose code:

```java
new NextcloudAPI(this, ssoAccount, new GsonBuilder().create(), new NextcloudAPI.ApiConnectedListener() {
    @Override
    public void onConnected() {
        Log.i(TAG, "SSO API connected for " + ssoAccount);
    }

    @Override
    public void onError(Exception e) {
        e.printStackTrace();
    }
});
```

The `onConnected` callback is usually quite useless, because according to the [README.md](https://github.com/nextcloud/Android-SingleSignOn#5-how-to-make-a-network-request) the requests are queued anyway.
The `onError` callback might be more interesting, to handle exceptions more gracefully, use own loggers, etc.

# Proposal

1. Add a default implementation in `NextcloudAPI.ApiConnectedListener#onConnected` which simply logs an info
2. Add a second constructor for `NextcloudAPI` without an `NextcloudAPI.ApiConnectedListener`. This can internally call the original constructor and add some default default behavior like `e.printStackTrace()` in the `onError` callback.

This would allow 3rd party apps to reduce the boilerplate code to:

```java
new NextcloudAPI(this, ssoAccount, new GsonBuilder().create(), Throwable::printStackTrace);
```

in the first case, and 

```java
new NextcloudAPI(this, ssoAccount, new GsonBuilder().create());
```

in the second, depending on the needs of the 3rd party app developer - while **still providing all features of today and maintaining full backward compatibility**.

David-Development:
@stefan-niedermann Sounds like a good idea to me 👍

PR #412, opened 2021-11-16
Update README.md samples

stefan-niedermann:
- Combine samples for `Activity` and `Fragment` - they are nearly identical and the duplication make everything quite complicated
- Update samples to use `var` keyword - makes it easier to focus on the actual stuff
- Update samples to use the less verbose constructor of `NextcloudAPI` without explicit `ApiConnectedListener` callback (see #400)
- Remove `<queries>` section. This can be omitted, since it is part of SSO lib `0.6.0` (see #282)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,42 +43,17 @@ repositories {
 }
 
 dependencies {
-    // Note: Starting with Version 0.5.7 Android Gradle Plugin (AGP) Version 7.0.0 or higher is
-    // required.
-    implementation "com.github.nextcloud:Android-SingleSignOn:0.5.7"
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
>     * Update samples to use `var` keyword - makes it easier to focus on the actual stuff

This only works for Kotlin, right? But all samples are in Java?

stefan-niedermann:
No, this is a [Java 10 feature](https://www.codejava.net/java-core/the-java-language/var-keyword-in-java) and we are using Java 11 🙂

tobiasKaminsky (approved):

Issue #26, opened 2018-08-27
Use LiveData in API


schaarsc:
this is more a question than an issue ...

I'm trying to become acquainted with this library and I have noticed that in the examples Observable is used as a return type. I was wondering why Observable was chosen instead of LiveData? LiveData is lifecycle aware, integrates nicely with ViewModel, Room, etc. and appears to be the recommended data type for data holders (https://developer.android.com/jetpack/docs/guide)

David-Dev:
![image](https://user-images.githubusercontent.com/3943392/44660698-c5cff280-aa08-11e8-9b3d-82c3939021e9.png)

desperateCoder:
We use LiveData in the Deck-app a lot. I have some methods written, that could come handy, when you are going to implement the support: https://github.com/stefan-niedermann/nextcloud-deck/blob/master/app/src/main/java/it/niedermann/nextcloud/deck/persistence/sync/adapters/db/util/LiveDataHelper.java

e.g. you can turn anything into a LiveData and stuff. Maybe this helps. If you have questions about it, you know how to contact me :wink:

stefan-niedermann:
AFAIK it is not possible to make network requests directly via `LiveData`. One would need to perform the actual `Call`s and map them into a new `MutableLiveData` instance - or transform an `Observable` into `LiveData`.

My understanding of a [good Android app architecture](https://developer.android.com/jetpack/guide#overview) is, that the `Repository` layer is responsible of delivering results from a local store (aka database) and synchronizing the contents of the database with unreliable remote resources in the background.
`LiveData` is therefore understandable needed to watch the data of a database via repository, but since the synchronization with remote sources should be handled in the background, there is no dependency to the current state of the `Activity` / `Fragment` here in my opinion.

@schaarsc are you aware of any app which is using `LiveData` in the `Remote Data Source` layer? Maybe i am missing something here, but i have never seen something like this in the wild yet.

However, i think a valid point is to make the current `SingleSignOnAccount` available as a `LiveData` which then can be used in combination with `Transformations#map` or `Transformations#switchMap`. I therefore created [a PR](https://github.com/nextcloud/Android-SingleSignOn/pull/415) which adds this feature and deprecates the usage of `(un)registerSharedPreferenceChangeListener`.

schaarsc:
I have not been active for quite some time, so I'm not sure how this would fit into the current state of implementations ...

The idea behind the original request was to handle all data with LiveData (local and remote).
[...]

PR #415, opened 2021-11-17
#26 Implement getCurrentSingleSignOnAccount() livecycle aware

stefan-niedermann:
- Add `CurrentSingleSignOnAccountLiveData` class which registers a `OnSharedPreferencesChanged` listener in a lifecycle aware way (It basically does only emit new events when the `Activity` / `Fragment` is not `paused` or `stopped`
- Add `SingleAccountHelper#getCurrentSingleSignOnAccount$` which returns a `CurrentSingleSignOnAccountLiveData` instance. The trailing `$` is a convention borrowed from `rxjs`[<sup>[1]</sup>](https://angular.io/guide/rx-library#naming-conventions-for-observables) to signalize that this is something observable
- `@Deprecate`ed `SingleAccountHelper#(un)registerSharedPreferenceChangeListener` methods, which are not lifecycle aware and potential memory leaks when used wrong
- Add sample usage to sample app

Example usage from 3rd party app:

```java
SingleAccountHelper.getCurrentSingleSignOnAccount$(this).observe(this, ssoAccount -> {
    Log.i(TAG, "New SingleSignOnAccount set: " + ssoAccount.name);
});
```


David-Development:
This looks great! I haven't personally worked with Live data before but definetely a great addition and perfect use case! I think we should probably mention it in the docs, eh?  Not sure if devs will find it if it isn't documented in the Readme..?

stefan-niedermann:
The PR is nust to demonstrate the general idea - of course we can add some docs and discuss the API before merging this. 🙂

stefan-niedermann:
I have added minimal documentation in the `README.md` and in the sample app :slightly_smiling_face:

stefan-niedermann:
Would love to get a review @AndyScherzinger @David-Development :slightly_smiling_face:

lib/src/main/java/com/nextcloud/android/sso/helper/SingleSignOnAccountLiveData.java
```diff
@@ -0,0 +1,74 @@
...
+    private SingleSignOnAccount getValueFromPreferences() throws NextcloudFilesAppAccountNotFoundException, NoCurrentAccountSelectedException {
+        return getCurrentSingleSignOnAccount(context);
+    }
+
+    @Override
+    protected void onInactive() {
+        sharedPrefs.unregisterOnSharedPreferenceChangeListener(preferenceChangeListener);
+        super.onInactive();
+    }
+}
```
AndyScherzinger, line 74: missing new line at the end of the file

stefan-niedermann:
More changes since the last review:

- In case of an error, the `LiveData<SingleSignOnAccount>` no emits <code>null</code> instead of nothing
- Added `@WorkerThread` annotations to synchronous methods
- Aligned method names (`setCurrentSingleSignOnAccount` → `applyCurrentSingleSignOnAccount` / `commitCurrentSingleSignOnAccount`), backwards compatible
- Added some more documentation

stefan-niedermann:
Actually, reading from `SharedPreferences` *is* synchronously. It ain't "heavy", but it is. This is the reason why the [`DataStore`](https://developer.android.com/topic/libraries/architecture/datastore) framework (on top of `SharedPreferences`) promotes the async operations and explicitly warns about using the synchronous APIs.

Issue #494, opened 2022-11-04
Migrate to Material 3 theme


stefan-niedermann:
Mainly the style of the dialogs has changed - more border radius and so on.

It's not a big deal, but looks quite inconsistent if the 3rd party app *and* the files app have new dialogs, but when importing an account the old dialogs are displayed...

cc @David-Development @AndyScherzinger @tobiasKaminsky

tobiasKaminsky:
@AlvaroBrey maybe this is a quick thing for you? :crossed_fingers:

Issue #470, opened 2022-07-29
If only dev app is installed SSO doesn't work


David-Development:
If users only have the dev version installed, the SSO connection doesn't work in the notes and news app.

According to the report in the notes app (see ticket below) I assume that it fails in the Token Exchange part - so the call to [authenticateSingleSignAccount](https://github.com/nextcloud/Android-SingleSignOn/blob/fe05d915ca689c590326b9ccdea1618140e0287c/lib/src/main/java/com/nextcloud/android/sso/AccountImporter.java#L336)

Ref:
- https://github.com/stefan-niedermann/nextcloud-notes/issues/1547
- https://github.com/nextcloud/news-android/issues/1033

David-Development:
@stefan-niedermann I figured out why it's not working anymore.. looks like that the dev app uses the production `accountType` ( `nextcloud` instead of `nextcloud.beta`). Therefore when we get the accounts from the Android Account Manager we have no idea where this account came from (dev, qa or beta). Since the dev app is using the same `accountType` as the production app we use the prod `packageId` to start the auth intent. In case the prod app is not installed the request will obviously fail. 

<img width="245" alt="Bildschirm­foto 2022-12-03 um 09 47 56" src="https://user-images.githubusercontent.com/4489723/205432620-5fb1a8f4-029e-436f-8e1f-e8ba1a36150f.png">

@tobiasKaminsky @AlvaroBrey 

After digging through the android files app source code it looks like that the logic on how to select the `accountType` hasn't been changed (last change was from Tobias 5 years ago) ([ref](https://github.com/nextcloud/android/blame/190f5dbbd36ec88d7a83053cdb46f345821f66e0/app/src/main/java/com/owncloud/android/MainApp.java#L636)). However I noticed a few inconsistencies - it looks like the dev resources were moved around inside the app back in march ([ref](https://github.com/nextcloud/android/blob/fbfa7fa746a509be165219d960030cf4a0967eea/app/src/versionDev/res/values/setup.xml)). I believe that the dev variables are **not** applied anymore. I can confirm, that my dev app doesn't have a beta indicator and the name of the app is just "Nextcloud" and not "Nextcloud Dev". I believe that the resources are not bundled in the app anymore - and I believe that the `build.gradle` is missing the resource sets for this flavor..?

I'm not sure where to find the QA apk but I assume that this might be an issue there as well.

tobiasKaminsky:
[...]

Issue #456, opened 2022-05-21
i18n: `Benötigt keine Übersetzung. Für Android wird nur die formelle Übersetzung verwendet (de_DE).`


stefan-niedermann:
Recently a user contacted me with the following error prompt in the Notes Android app (LineageOS 18.1):

![image](https://user-images.githubusercontent.com/4741199/169645851-a5f7d42f-72ce-4a20-9f65-19680c19c744.png)

Separate from being guilty myself for the failure itself, I was curious about the error message, so I did some research.

There are two german translations: [`de`](https://github.com/nextcloud/Android-SingleSignOn/blob/master/lib/src/main/res/values-de/strings.xml) and [`de-rDE`](https://github.com/nextcloud/Android-SingleSignOn/blob/master/lib/src/main/res/values-de-rDE/strings.xml). While the translations of `de-rDE` look fine, all translations of `de` look quite odd.

Probably one thought, they would not ever appear on the UI, but this seems to be the case.

- Should one dig more into this?
- Should we simply add proper traanslations to the `de` file`?

David-Development:
Hm.. weird. I assume the user is in austria or switzerland? I believe for germany the language code would be `de-rDE`. Since we don't have `de-rCH` the fallback will be `values-de`. At leasts that's what I'm getting from the following explanation on stackoverflow. So we'd only need `values-de` and `values-de-rDE` can and probably should be deleted if we don't plan on having any special translations for Germany / Switzerland / Austria.

> The difference between values-de-rDE and values-de is that the first is referring to the country and language, while the other only requires the language.
>
>What region is chosen can be determined in System Settings. German (Germany) is values-de-rDE while German (Switzerland) is values-de-rCH, but both are values-de.
>
>If the system is set to German (Germany), it will first read all values from values-de-rDE, then the missing values from values-de, and finally from the default values.
>
>This is useful for when values contains American English, and values-en-rGB can provide only the strings that are different (i.e. colour, flavour, customise).

Source: https://stackoverflow.com/a/26415196/13370504

stefan-niedermann:
So what needs to be done to fix this? Just delete the region derivates? Ensure they aren't transferred from transifex in the future?

David-Development:
@stefan-niedermann Exactly. At least that's my understanding 😅 Just keep `values-de` and delete all the other derivates.

stefan-niedermann:
How do we ensure they won't be readded by the next transifex merge?

[...]

PR #495, opened 2022-11-07
Use material 3 for error dialogs and sample app

AlvaroBrey:
Related: #494

| Before | After |
| --- | --- |
| Screenshot_1667823171 | Screenshot_1667823131 |

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -105,6 +105,8 @@ dependencies {
     implementation 'androidx.core:core:1.8.0'
     implementation 'androidx.fragment:fragment:1.5.4'
 
+    implementation 'com.google.android.material:material:1.6.1'
+
     compileOnly "org.projectlombok:lombok:1.18.24"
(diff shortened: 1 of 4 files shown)


David-Development:
Looks great to me! So much better than the old design 😅 @tobiasKaminsky @stefan-niedermann  Is this something we should publish in a new version right away?

stefan-niedermann:
>  Is this something we should publish in a new version right away?

Actually I'd like to get a fix for #456 and #470 in first before releasing a new version cause I have quite a few users which regularly complain about those two issues :see_no_evil:

stefan-niedermann (approved):
Thanks for the awesome work @AlvaroBrey !

David-Development:
@stefan-niedermann Good point about #470. Not sure when I'll have time to work on it.. I'm drowning in work right now and it doesn't look like it's getting any better in the next 2 weeks at least - I'll create a reminder though. If you already start investigating it let me know

PR #528, opened 2023-03-06
Support rxjava3 return type

p1gp1g:
This add support for rxjava3 (to do this: https://github.com/UP-NextPush/android/commit/773862c7fe71cab7404406e08c74403d7bfe1c6f)

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -113,6 +113,7 @@ dependencies {
     api 'com.google.code.gson:gson:2.10.1'
 
     implementation 'io.reactivex.rxjava2:rxjava:2.2.21'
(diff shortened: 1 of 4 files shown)


David-Development:
@p1gp1g Thank you for adding support for rxjava3 - looks good to me! @stefan-niedermann @tobiasKaminsky any objections? 

Reminder to myself.. probably a good idea to migrate the news app to rxjava3 as well.. 😅

David-Development (approved):
min-sdk level for rxjava3 is 21+ (same as Single Sign On Library right now). So all good ✅

p1gp1g:
I've just seen that it misses a change I haven't pushed

p1gp1g:
Here it is

Issue #541, opened 2023-04-19
Instantiating `Void` does no longer work with AGP 8


stefan-niedermann:
We implemented [a hacky way to instantiate `Void` using reflection](https://github.com/nextcloud/Android-SingleSignOn/commit/e19aae1cc0f124cc58f9dbc726413ddb6de7ffc6). This does no longer work with new Java versions and encumbers users while upgrading to AGP 8 and newer Java versions.

Why did we implement this at all? See [comment](https://github.com/nextcloud/Android-SingleSignOn/issues/541#issuecomment-1893453525), TLDR; Using REST endpoints that return an empty body as `Observable<Void>` leads to a crash. (See https://github.com/nextcloud/Android-SingleSignOn/issues/231).

CC @desperateCoder @David-Development @AndyScherzinger @tobiasKaminsky

stefan-niedermann:
One possible solution I could imagine is to provide a new type like `EmptyResponse` from the SSO lib, which could be used in the 3rd party apps to declare this kind of type. Of course it would need to be documented to use this type instead of `Void`...

stefan-niedermann:
Reminder that updating ProGuard rules *might* be necessary after fixing this: https://github.com/nextcloud/Android-SingleSignOn/pull/545/files

stefan-niedermann:
FYI: I am currently trying to make the problem reproducible in the `sample` app on the `agp-8` branch, but it still works there... 🤔

AndyScherzinger:
Are all parts compiling targeting java17?

stefan-niedermann:
Hehe, **Unit tests** fail on `agp-8`, the sample app works - I can only assume that the Java runtime for Android does not (yet) ship the protected modules. So: It is reproducible now with the `agp-8` branch when running the unit tests.

AndyScherzinger:
@stefan-niedermann looks good to me. How would I need to test this?

stefan-niedermann:
A bit complicated! My use case is in the Deck Android app (we need REST endpoints with no return body), but it uses SSO as direct and transitive dependency.

Maybe utilizing the SSO sample app and try to request a Deck endpoint with `Observable` and / or `Call` response types?

As you see, this PR was coded blindly, testing is quite hard and help is appreciated.

AndyScherzinger:
Are you aware of any of those used by the notes app maybe? Than I could give them a try

stefan-niedermann:
[...]

PR #542, opened 2023-04-19
fix(void): Introduce new EmptyResponse type as successor for Void (#541)

stefan-niedermann:
:warning: **BREAKING CHANGE** :warning: 

- Fix for #541
- Can be tested [as described here](https://github.com/nextcloud/Android-SingleSignOn/issues/541#issuecomment-1893453525)


David-Development:
Looks like a promising solution to the Void issue with AGP-8! I'm not really sure how to test it in the Okhttp3Helper .. does anyone actually use the Okhttp3Helper?

stefan-niedermann:
@AndyScherzinger @tobiasKaminsky @David-Development can you please review the idea and check whether everything still works as expected?

This issue does prohibit 3rd party apps using recent build tools and I'd like to get it solved. Other opinions or ideas are welcome of course.

David-Development:
@stefan-niedermann I created a few unit tests to cover the EmptyResponse check. I think the Nextcloud Files app would have to send the `EmptyResponse` as well to make it work properly (currently the unit test mocks the files api and returns a `EmptyResponse`). I think returning `Void` as an Observable is an edge case anyways where I don't even see any use-cases for. I'd say we can just merge this change (with a major version update) and keep an eye out if any of the third party apps break (and of course mention it as a breaking change).

My tests can be found here: https://github.com/nextcloud/Android-SingleSignOn/pull/589

stefan-niedermann:
> I think the Nextcloud Files app would have to send the EmptyResponse as well 

This implies that the Files app does today send `Void` instances? How does this work? Just sending `null`?

> I think returning Void as an Observable is an edge case anyways where I don't even see any use-cases for. 

Well, the Deck Android App uses `Observable<Void>` as return type quite a lot (though @desperateCoder  is [working on replacing `Observable`s by `Call`s](https://github.com/stefan-niedermann/nextcloud-deck/tree/observable2call)).

> I'd say we can just merge this change (with a major version update) and keep an eye out if any of the third party apps break (and of course mention it as a breaking change).

Hm, What exactly would we need to adjust in the Files app to make this scenario work properly?

David-Development:
@stefan-niedermann 
> This implies that the Files app does today send Void instances? How does this work? Just sending null?

I'm not 100% sure anymore since it's been a while but I looked at the code again and the files app doesn't know anything about types. So I think it we don't need any changes there..  ([ref](https://github.com/nextcloud/android/blob/master/app/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java)). But I'm not sure anymore what exactly it will send back to our app in case we don't have a response.

If you are already using the `Observable<Void>` in the Deck app maybe you can test it there? And see what kind of response you are getting from the Files app? Or if the code already works as is.

David-Development (approved):

tobiasKaminsky (approved):

PR #543, opened 2023-04-19
add response headers to responses for `Call<T>` Retrofit methods

desperateCoder:
For now, only the `Observable<T>` calls can get response headers.

We (Nextcloud Deck for Android) wanted to get rid of `RxJava` in our project, so we migrated to `Call<T>` provided by Retrofit.

In this case, the Headers weren't included in the responses.

CC: @David-Development @tobiasKaminsky


stefan-niedermann (approved):

stefan-niedermann:
Tested in real live project, works like a charm.

Issue #488, opened 2022-10-05
Break on minified app


p1gp1g:
When minify is turned on, the SSO doesn't work anymore (cf. https://github.com/UP-NextPush/android/commit/e3ff269a1f85556732f976f5b606b1a545a06b30)

```
> adb logcat --pid=(adb shell pidof -s com.nextcloud.client)


10-05 20:39:26.040  7489  7515 E InputStreamBinder: Error during Nextcloud request
10-05 20:39:26.040  7489  7515 E InputStreamBinder: java.lang.ClassNotFoundException: h3.c
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.lang.Class.classForName(Native Method)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.lang.Class.forName(Class.java:454)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.io.ObjectInputStream.resolveClass(ObjectInputStream.java:683)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:1731)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1622)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1900)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1440)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at java.io.ObjectInputStream.readObject(ObjectInputStream.java:428)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at com.nextcloud.android.sso.InputStreamBinder.deserializeObjectAndCloseStream(InputStreamBinder.java:226)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at com.nextcloud.android.sso.InputStreamBinder.performNextcloudRequestAndBodyStreamV2(InputStreamBinder.java:127)
10-05 20:39:26.040  7489  7515 E InputStreamBinder:     at com.nextcloud.android.sso.InputStreamBinder.performNextcloudRequestV2(InputStreamBinder.java:111)
[...]

PR #545, opened 2023-04-19
Add consumer proguard rules

Unpublished:
As requested in #488

Any feedback is welcome especially on `README.md` section!


stefan-niedermann (approved):
Approving as far as I can tell - I am not a ProGuard Pro :wink: but the documentation is understandable and the rules are definitely better than no rules at all - which are known to be broken.

PR #546, opened 2023-04-21
feat: Implement `java.lang.AutoClosable`

stefan-niedermann:
#544 inspired me to implement the `AutoClosable` interface for the whole library.
It allows 3rd party apps a graceful release of resources even in case of an exception.
It is also (still) backward compatible.


David-Development (approved):
Great idea!

AndyScherzinger (approved):
Nice improvemt 👍
I like it!

PR #556, opened 2023-05-17
feat: Remove outdated icon from error dialog

stefan-niedermann:
The default :warning: icon (`android.R.drawable.ic_dialog_alert`) looks extremely outdated and suffers from low contrast.

In this PR the icon simply gets removed.

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/lib/src/main/java/com/nextcloud/android/sso/ui/UiExceptionManager.java
+++ b/lib/src/main/java/com/nextcloud/android/sso/ui/UiExceptionManager.java
@@ -11,7 +11,6 @@ import android.text.method.LinkMovementMethod;
 import android.text.util.Linkify;
 import android.widget.TextView;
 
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
@stefan-niedermann couldn't we ship the icon? Also common has theming support for material alert dialogs also theming the icon according to m3 and then we should also be just fine.

But I am also fine with just removing it. Both would fit the m3 specs 👍

stefan-niedermann:
Regarding the icon: I tend to just removing it...

Regarding the theming support: yup, sounds like a good idea to and `android-common`, though this will probably be more effort, because afaik only the content aread of an `AlertDialog` can be themed (not the title and not the buttons), so we would need to implement those parts ourself. Correct me if I'm wrong. However, this should be done independently in a separate branch

AndyScherzinger (approved):

Issue #557, opened 2023-05-17
Improve SSO error dialogs shown by 3rd-party-apps


stefan-niedermann:
Current | Proposal
--- | ---
![grafik](https://github.com/nextcloud/Android-SingleSignOn/assets/4741199/28ff4027-6e13-46b5-b0a8-728ccf747597) | ![sample](https://github.com/nextcloud/Android-SingleSignOn/assets/4741199/ec457604-0c39-4887-8177-630298bcaf67)

# Primary action

## Current situation

Currently each error dialog just has an <kbd>Ok</kbd> button, some dialogs do have links in their description (see screenshot above)

## Proposal

Extend `SSOException` by two further (optional) fields:

- `@StringRes String primaryActionText`
- `Intent primaryAction`

When generating the dialog, we can check them and display a dialog like the one above with an <kbd>Install</kbd> button instead of a link in the description. Additionally, if a `primaryAction` is available, we should display a <kbd>Close</kbd> button next to the primary action.

# Messages

Especially the titles don't look that good. Maybe we can improve the wording here

```xml
<string name="no_current_account_selected_exception_title">Warning</string>
<string name="no_current_account_selected_exception_message">No account selected yet</string>

<string name="android_get_accounts_permission_not_granted_exception_title">Error</string>
<string name="android_get_accounts_permission_not_granted_exception_message">In order to use the Nextcloud Single-Sign-On on this device, you need to grant permission to access your accounts</string>

<string name="current_account_not_found_exception_title">Warning</string>
<string name="current_account_not_found_exception_message">Selected account was not found. Something went wrong.</string>

<string name="token_mismatch_title">Error</string>
<string name="token_mismatch_message">Provided authentication token does not match or package name is not permitted to perform this action</string>

<string name="nextcloud_files_app_not_installed_title">Error</string>
<string name="nextcloud_files_app_not_installed_message" formatted="true">Nextcloud Files app is not installed. Please install it: %1$s</string>

<string name="nextcloud_files_app_account_not_found_title">Error</string>
<string name="nextcloud_files_app_account_not_found_message">Account not found in Nextcloud Files app</string>

<string name="nextcloud_files_app_account_permission_not_granted_title">Error</string>
<string name="nextcloud_files_app_account_permission_not_granted_message">You have to accept the requested permissions to use the single sign-on feature.</string>

<string name="nextcloud_http_request_failed_title">Error</string>
[...]

PR #558, opened 2023-05-17
Enhance exception handling

stefan-niedermann:
Note: This PR contains also the changes of https://github.com/nextcloud/Android-SingleSignOn/pull/556

Fixes: https://github.com/nextcloud/Android-SingleSignOn/issues/557

@nextcloud/designers  @nimishavijay @jancborchardt please review the [:arrow_right: new wording of the error titles and messages :arrow_left:](https://github.com/nextcloud/Android-SingleSignOn/pull/558/files#diff-2a7fd72ed64345780392ca16d959ede38bc4180bacb0ac1c6b19234cd5b94fef).

@AndyScherzinger @tobiasKaminsky @David-Development 

Sorry that the PR got a bit bigger than expected, but I wanted to get rid of this reflection access in `SSOException`, which I considered as a technical dept:

```java
Class.forName("android.app.ActivityThread").getMethod("currentApplication").invoke(null, (Object[]) null)
```

I therefore refactored the construction of the exceptions which is likely a breaking change (though I tried to maintain backward compatibility where possible).

Guess this needs a careful review.


lib/src/main/res/values/strings.xml
```diff
@@ -1,44 +1,53 @@
...
+    <string name="nextcloud_http_request_failed_title">Request failed</string>
     <string name="nextcloud_http_request_failed_message" formatted="true">HTTP request failed with HTTP status-code: %1$d</string>
 
     <string name="nextcloud_unsupported_method_title" translatable="false">Error</string>
-    <string name="nextcloud_unsupported_method_message" translatable="false">The requested HTTP Method is not supported yet! Please contact the developers of the Android Single-On Library for Android</string>
+    <string name="nextcloud_unsupported_method_message" translatable="false">The requested HTTP Method is not supported yet! Please contact the developers of the Single Sign On Library for Android</string>
+    <string name="nextcloud_unsupported_method_action">Contact</string>
 
-    <string name="nextcloud_invalid_request_url_title" translatable="false">Error</string>
+    <string name="nextcloud_invalid_request_url_title" translatable="false">Invalid URL</string>
```
jancborchardt, line 35: We generally say "link" instead of "URL" if it's user-facing. Or is this for devs?

lib/src/main/res/values/strings.xml
```diff
@@ -1,44 +1,53 @@
...
     <string name="nextcloud_invalid_request_url_message" translatable="false" formatted="true">The requested URL is invalid: %1$s</string>
 
-    <string name="nextcloud_files_app_not_supported_title">Error</string>
-    <string name="nextcloud_files_app_not_supported_message" formatted="true">Your Nextcloud Files app currently does not support the single sign on feature. Please update it: %1$s</string>
+    <string name="nextcloud_files_app_not_supported_title">Update Nextcloud Files app</string>
+    <string name="nextcloud_files_app_not_supported_message">Your Nextcloud Files app is outdated and does not yet support the Single Sign On feature.</string>
```
jancborchardt, line 39: Better would be: "Your Nextcloud Files app is an older version which does not yet support the Single Sign On feature."

"Outdated" sounds like it's generally bad, older version is more specific.

jancborchardt (changes requested):
Seems fine, just 2 comments inline. :)

lib/src/main/res/values/strings.xml
stefan-niedermann: Yeah, it's more directed to devs. It's not a link which a user is supposed to open in the browser (like `cloud.nextcloud.com`).
It's an URL to an API endpoint, so I'd rather keep the term `URL` here.

lib/src/main/res/values/strings.xml
stefan-niedermann: Thanks, updated!

AndyScherzinger (approved):

AndyScherzinger (approved):

jancborchardt (approved):

David-Development:
@stefan-niedermann Much cleaner UI, I love it! Looks great to me! I think I went down the reflection route because I was afraid of causing memory leaks by leaking the context in another class. But your solution looks great! There are no references to the Exceptions after it has been sent to the client app hence there shouldn't be any memory leaks.

David-Development (approved):

PR #561, opened 2023-05-21
enable aidl buildFeature

David-Development:
build fails on latest Android Studio if the buildFeature is not explicitly enabled

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -34,6 +34,10 @@ android {
         consumerProguardFiles 'consumer-proguard-rules.pro'
     }
 
+    buildFeatures {
+        aidl true
+    }
+
     buildTypes {
         release {
             minifyEnabled false


stefan-niedermann:
> latest Android Studio

Which version are you using?

I can't reproduce with

```
Android Studio Flamingo | 2022.2.1 Patch 1
Build #AI-222.4459.24.2221.9971841, built on April 19, 2023
```

> Setting the value to `null` resets to the default value. Default value is `true`.
  Source: [developer.android.com](https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/dsl/BuildFeatures#aidl)

David-Development:
@stefan-niedermann I'm using the exact same version. Maybe it has something to do with the fact that the News App is already using AGP 8. I'm testing it in conjunction with the News Android App. Maybe it will be fixed automatically once this lib uses AGP 8 as well

stefan-niedermann (approved):

PR #562, opened 2023-05-21
enh: Use `try-with-resources` for instantiating ExponentialBackoff

stefan-niedermann:
Interestingly this does break the whole SSO and I can't explain why. Maybe someone wants to have a look at this?


David-Development:
@stefan-niedermann What breaks exactly? The new Version 0.8.0 runs perfectly find with the Nextcloud News App.

stefan-niedermann:
0.8.0 and master are fine. But the changes on this branch are breaking the lib and I don't understand why...

stefan-niedermann:
Ah, this PR is just dumb. Of course everything break because after the `try` block the connection gets closed, so it can't work 🤦‍♂️

PR #567, opened 2023-06-04
fix crash if no store is installed on the phone

David-Development:
Handles the case that if the files app not installed dialog is shown and no store is installed the app crashes.

Instead it will show an additional dialog not telling the user to install the app manually by redirecting the user to the Nextcloud Clients website.

https://github.com/nextcloud/Android-SingleSignOn/assets/4489723/a46a3fcd-7d37-4116-8256-4bbe01d97357


lib/src/main/java/com/nextcloud/android/sso/exceptions/NoStoreInstalledException.java
```diff
@@ -0,0 +1,56 @@
...
+    public NoStoreInstalledException(@NonNull Context context, @Nullable Throwable cause) {
+        super(
+                context.getString(R.string.nextcloud_files_app_no_store_installed_message),
+                R.string.nextcloud_files_app_no_store_installed_title,
+                cause
+        );
+    }
+
+    public Optional<Intent> getPrimaryAction() {
+        String url = "https://nextcloud.com/install/#install-clients";
```
stefan-niedermann, line 47: Please put this URL into `strings.xml` next to the other URLs

lib/src/main/java/com/nextcloud/android/sso/ui/UiExceptionManager.java
```diff
@@ -47,7 +50,21 @@ public static void showDialogForException(@NonNull Context context,
         final var optionalAction = exception.getPrimaryAction();
 
         if (optionalAction.isPresent()) {
-            showDialogForException(context, exception, actionText, (dialog, which) -> context.startActivity(optionalAction.get()));
+            showDialogForException(context, exception, actionText, (dialog, which) -> {
+                try {
+                    context.startActivity(optionalAction.get());
+                } catch(ActivityNotFoundException anf) {
```
stefan-niedermann, line 56: Line formatting seems a bit off? <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>L</kbd>

lib/src/main/java/com/nextcloud/android/sso/ui/UiExceptionManager.java
```diff
@@ -47,7 +50,21 @@ public static void showDialogForException(@NonNull Context context,
         final var optionalAction = exception.getPrimaryAction();
 
         if (optionalAction.isPresent()) {
-            showDialogForException(context, exception, actionText, (dialog, which) -> context.startActivity(optionalAction.get()));
+            showDialogForException(context, exception, actionText, (dialog, which) -> {
+                try {
```
stefan-niedermann, line 54: I am wondering whether we can check for an existing store *before* trying to launch it? But as I already said: It's an edge case so just reacting to the exception is fine for me, too.

stefan-niedermann (changes requested):

lib/src/main/java/com/nextcloud/android/sso/ui/UiExceptionManager.java
David-Development: Probably "easier" and safer this way.. Otherwise we might get issues in case new stores pop up and we need to adjust the intents etc. And I believe we would need to register the "query" permission to even know about each store..

stefan-niedermann: It seems to be [possible](https://stackoverflow.com/questions/15407502/how-to-check-if-an-intent-can-be-handled-from-some-activity) without querying specific packages:

```java
intent.resolveActivity(getPackageManager()) != null;
```

So maybe we don't even need another exception type, but can just return a different `Intent` in `NextcloudFilesAppNotInstalledException`?

stefan-niedermann: @David-Development Please allow me to propose some changes for this PR: https://github.com/nextcloud/Android-SingleSignOn/pull/568

It will directly open the "correct" error dialog if no store is present.

stefan-niedermann:
We probably should do a `0.8.1` bugfix release after merging this...

PR #568, opened 2023-06-06
Refactor no-store-installed handling

stefan-niedermann:
Just a proposal :)


David-Development:
Looks great to me! I like this better than my proposal!

stefan-niedermann:
Yeah, actually a static factory method would be even better to avoid the redundant checks but it's not that bad because

- it are only boolean checks
- they only get checked in case of an error, not always. When we need to handle an error, performance is not the highest goal 😉

PR #635, opened 2024-01-20
Update R8/ProGuard section of README.md

### LINKED ISSUE #572: Readme contains bad R8 advice
> R8 and ProGuard rules are bundled into [SSO](https://github.com/nextcloud/Android-SingleSignOn/blob/master/lib/consumer-proguard-rules.pro). The bundled rules do not cover enabled obfuscation. Therefore it is recommended to add `-dontobfuscate` to your app-specific proguard rules.

Setting `-dontobfuscate` on a whole app is a bad idea, as it will unnecessarily increase the app size.

There are exactly two possibilities:
1) This library works in a consuming app that has R8 fullmode and obfuscation turned on. In this case, the text above is wrong and should be changed.

2) The library does not work in a consuming app that has R8 fullmode and obfuscation turned on. In this case, the R8 rules provided by the library should be improved until it does. Offloading this work to consuming developers is lazy and bad practice.

As fas as I can tell we are dealing with 1) here.

So I suggest altering the Readme to say the following:

The rules needed to make SSO work with R8 and ProGuard are [bundled into the library](https://github.com/nextcloud/Android-SingleSignOn/blob/master/lib/consumer-proguard-rules.pro). 

With [R8 full mode](https://r8.googlesource.com/r8/+/refs/heads/master/compatibility-faq.md#r8-full-mode) being enabled by default since [AGP 8.0](https://developer.android.com/build/releases/gradle-plugin#default-changes), you will probably need to handle following app-specific rules yourself (or disable full mode):
...

(Although I don't understand why you would want to repeat Gson/Retrofit rules, if devs have a problem with them they can go to the respective docs themselves)

connyduck:
closes #572

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,11 +281,9 @@ if (VersionCheckHelper.verifyMinVersion(context, MIN_NEXTCLOUD_FILES_APP_VERSION
 
 ## R8/ProGuard
 
-R8 and ProGuard rules are bundled into [SSO](lib/consumer-proguard-rules.pro).
-The bundled rules do **not** cover enabled obfuscation.
-Therefore it is **recommended** to add `-dontobfuscate` to your app-specific proguard rules.
+R8 and ProGuard rules are bundled into [SSO](lib/consumer-proguard-rules.pro) so it will automatically work without additional app level rules.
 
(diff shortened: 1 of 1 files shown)


connyduck:
Are there known breakages in downstream apps? For the two apps I worked with it was fine.

connyduck:
I built the [news-android](https://github.com/nextcloud/news-android) app without `-dontobfuscate`. The resulting apk is about 7% smaller (4965kB vs 5319kB) and works just fine. I don't see any rules I'd call too broad.

connyduck:
Oh that are the application classes, I was looking for `com.nextcloud`, my bad. Here is a suggestion to improve the situation:
https://github.com/nextcloud/news-android/pull/1382/

PR #636, opened 2024-01-21
add migration info for Void -> EmptyResponse

David-Development:


stefan-niedermann:
Hmm... I can't reproduce this issue 🧐 We are using `Call<Void>` successfully in the Deck app (at least on a branch).

> Could not instantiate java.lang.Void, because response was null.

This error message suggests one is trying to instantiate Void (the workaround we removed in SSO 1.0.0) and it must not happen any longer. Are you *absolutely* sure you do not use SSO below 1.0.0 and also don't have idk any transitive dependencies or so?

David-Development:
I cloned the Single-Sign-On Repo into the news app and included it as a gradle project dependency in the settings.gradle. So I'm fairly sure that it's version 1.0.0 🙈 And in the deck app it works with the latest version and the `Call<Void>`?

stefan-niedermann:
On the `observable2call` branch, yup :shrug: 

Where exactly is the error message (`Could not instantiate java.lang.Void, because response was null.`) coming from? At some point `Void` must get instantiated - where is it?

David-Development:
@stefan-niedermann Here: https://github.com/nextcloud/Android-SingleSignOn/pull/636/files#diff-c367b4028a9db1f826b9e7657605ad06e197238f3ab5e4a6d438dcd8c01d1983R151

The response is null (because the server doesn't return anything) so it checks if the the  `targetEntity` is an Object - which it isn't so it throws the exception. Maybe in your case the server returns something and you just ignore it? That would probably be a different case

lib/src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
```diff
@@ -141,7 +141,10 @@ public <T> T convertStreamToTargetEntity(InputStream inputStream, Type targetEnt
             } else {
                 result = gson.fromJson(targetReader, targetEntity);
                 if (result == null) {
-                    if (targetEntity == Object.class) {
+                    if(targetEntity == Void.class) {
+                        throw new IllegalStateException("Detected Void return type - please " +
+                                "switch to the new return type \"EmptyResponse\" instead.");
```
stefan-niedermann, line 146:
```suggestion
                        ensureTypeNotVoid(Void.class);
```

stefan-niedermann: Suggesting to reuse the same method (looks silly but ensures the same error message)

lib/src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java
```diff
@@ -141,7 +141,10 @@ public <T> T convertStreamToTargetEntity(InputStream inputStream, Type targetEnt
...
+                    } else if (targetEntity == Object.class) {
```
stefan-niedermann, line 147: Can't we use 

```java
ensureTypeNotVoid(targetEntity);
```

to ensure the same error message?

David-Development: Yes, good point! 👍

stefan-niedermann: Actually, if we call `ensureTypeNotVoid` in `performNetworkRequestV2` instead of all the current places, we should easily tackle all cases, no?

David-Development: In theory - yes. But `performNetworkRequestV2` has no response type information thus we can't add the check there.

stefan-niedermann (approved):

stefan-niedermann (changes requested):
One more thing: Please adjust the README, there's a hint to `EmptyResponse` which should be rewritten to also match `Call`s then

stefan-niedermann (approved):
Not sure whether the explanation is still valid (given `Call`s are not related to `RxJava`), but yeah, the hint is there.

PR #735, opened 2024-07-25
Dynamic SSO

tobiasKaminsky:
with 
SingleSignOn.getSingleSignOn().init(new FilesAppType("com.nextcloud.cloud.android.client", "com.nextcloud.cloud"));

another Files app can be requested.

### DIFF
--- a/lib/src/main/java/com/nextcloud/android/sso/model/FilesAppType.java
+++ b/lib/src/main/java/com/nextcloud/android/sso/model/FilesAppType.java
@@ -8,33 +8,19 @@
 package com.nextcloud.android.sso.model;
 
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 
-public enum FilesAppType {
-
-    PROD("com.nextcloud.client", "nextcloud"),
-    QA("com.nextcloud.android.qa", "nextcloud.qa"),
-    DEV("com.nextcloud.android.beta", "nextcloud.beta");
-
+public class FilesAppType {
     public final String packageId;
     public final String accountType;
+    public final Type type;
 
-    FilesAppType(@NonNull String packageId, @NonNull String accountType) {
+    public FilesAppType(@NonNull String packageId, @NonNull String accountType, Type type) {
         this.packageId = packageId;
         this.accountType = accountType;
+        this.type = type;
     }
 
-    /**
-     * @return {@link #PROD}, {@link #QA} or {@link #DEV} depending on {@param accountType}.
-     * Uses {@link #PROD} as fallback.
-     */
-    @NonNull
-    public static FilesAppType findByAccountType(@Nullable String accountType) {
-        for (final var appType : FilesAppType.values()) {
-            if (appType.accountType.equalsIgnoreCase(accountType)) {
-                return appType;
-            }
-        }
-        return PROD;
+    public enum Type {
+        PROD, QA, DEV
     }
 }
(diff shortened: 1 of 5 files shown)


lib/src/main/java/com/nextcloud/android/sso/SingleSignOn.java
```diff
@@ -0,0 +1,52 @@
...
+     * Uses {@link #PROD} as fallback.
+     */
+    @NonNull
+    public FilesAppType findByAccountType(@Nullable String accountType) {
+        for (final var type : types) {
+            if (type.accountType.equalsIgnoreCase(accountType)) {
+                return type;
+            }
+        }
+        return FilesAppType.PROD;
```
stefan-niedermann, line 50: Careful here, if `FilesAppType` is not included in the `types` list, throwing an exception might be better to prevent an invalid state, no? Also the Javadoc comment should be updated.

lib/src/main/java/com/nextcloud/android/sso/SingleSignOn.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+public class SingleSignOn {
+    private static final SingleSignOn SINGLE_SIGN_ON = new SingleSignOn();
+    private List<FilesAppType> types = new ArrayList<>();
```
stefan-niedermann, line 13:
```suggestion
    private final List<FilesAppType> types = new ArrayList<>(3);
```

lib/src/main/java/com/nextcloud/android/sso/model/FilesAppType.java
```diff
@@ -8,33 +8,19 @@
...
 
-public enum FilesAppType {
+public class FilesAppType {
 
-    PROD("com.nextcloud.client", "nextcloud"),
-    QA("com.nextcloud.android.qa", "nextcloud.qa"),
-    DEV("com.nextcloud.android.beta", "nextcloud.beta");
+    public static FilesAppType PROD = new FilesAppType("com.nextcloud.client", "nextcloud");
+    public static FilesAppType QA = new FilesAppType("com.nextcloud.android.qa", "nextcloud.qa");
+    public static FilesAppType DEV = new FilesAppType("com.nextcloud.android.beta", "nextcloud.beta");
```
stefan-niedermann, line 16: Please ensure, that those constants are not referenced directly within the SSO lib anywhere else. I think there might be one or two positions where we fallback to `PROD` or anything else in case we don't find the actual type. This is an mistake after this PR gets merged because it is an assumption that might be wrong when people can register or unregister types as they want.

lib/src/main/java/com/nextcloud/android/sso/SingleSignOn.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.nextcloud.android.sso.model.FilesAppType;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+public class SingleSignOn {
```
stefan-niedermann, line 11: I don't like the name as it is kind of too generic. Maybe `FilesAppTypeRegistry` or anything else more verbose would be helpful?

stefan-niedermann (approved):

PR #759, opened 2024-09-29
chore(records)!: Migrate data classes to records

stefan-niedermann:
This is a (small and easy to adapt) breaking change, but records are the structures intended to be used for pure data objects.

Note: The `FilesAppTypeRegistry` itself introduced in `1.3.0` was a breaking change, too.


AndyScherzinger:
Well, I'd it is fine to do and also we can simply bump to v2.0.0 to comply with semantic versioning. It is a library after all, so it is also fine to just move forward and do such changes. The number of apps making use of the lib is still relatively small and if the breaking change is easy to adapt to, that is even better then 👍

AndyScherzinger (approved):

PR #760, opened 2024-09-29
chore(exception)!: Enhance exception handling

stefan-niedermann:
- Replace http://localhost/ with http://example.com[/] because localhost is a possible valid host and therefore confusing for users. Considered info-removed-for-privacy-reasons.com (or something similar), but there is no guarantee that this domain is not registered. example.com can by definition not be registered.
- Unify way to create a response on an exception by passing the whole exception object into the helper method
- Use HTTP 900 for unknown exceptions to distinguish other proprietary status codes: HTTP 520 is for example used by Cloudflare. Note that HTTP 900 is also not standardized.


AndyScherzinger (approved):
LGTM, codacy can be ignored 👍

David-Development (approved):
Great changes!

PR #779, opened 2024-11-11
Remove desugaring

p1gp1g:
Desugaring allows to use new Java features and SDK API which are usually embedded in the OS to older device and Java version. This works by embedding these features in the apk.

This library uses it for SDK<24.

A project using this library must enable desugaring, and keep up to date the desugaring library with this project and their AGP. It makes the set up process and maintain process a bit harder.

If I'm correct, it is possible to allow project to enable desugaring if they need SDK>=24 by removing this compile option from the library and let it be done by projects. This sounds like a good idea but this may be counter productive, for instance some project prefer to drop support for older SDK than doing this (I took the first example I've seen, for another project: https://github.com/elastic/apm-agent-android/pull/59).
It would also silently fail for project targeting lower SDK that forget to set this compile option.

This is probably why desugaring is not common in libraries

If we look closer, this has been enable for a small set of features, that are easy to replace:
- var (Java 11), easily replaced by the explicit class
- yield (Java 14), used as a throw
- record (Java 14), easily replaced with `public final` fields and let the IDE write the constructor
- Streams (SDK>=24)
- Optional (SDK>=24)

So, this is a suggestion to get rid of these few features. As a counter side, this leads to a few breaking changes:
- previous record fields aren't access from Java project the same way: from `appType.packageId()` to `appType.packageId`
- lib/src/main/java/com/nextcloud/android/sso/exceptions/SSOException.java functions return nullable type instead of Optional\<T>

So, because of the existence of breaking changes, there are pro/cons to keep or remove desugaring.

[Edit: clarify drawbacks]

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -55,9 +55,8 @@ android {
     }
 
     compileOptions {
-        coreLibraryDesugaringEnabled true
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
     }
 
     publishing {
@@ -114,7 +113,6 @@ detekt {
 }
 
 dependencies {
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.3'
 
     implementation "androidx.appcompat:appcompat:1.7.0"
     implementation 'androidx.annotation:annotation:1.9.1'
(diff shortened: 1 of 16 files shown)


stefan-niedermann:
> equals and hashCode aren't used here isn't it ?

You make assumptions about other third party apps, that may rely on a working comparison or sorting or backpressure detection - all features that rely on a proper `equals` method.

I still don't get why anyone should *not* want to use new language features that are backward compatible for the users.
Putting 2 lines of "code" (in config files) to allow all of those features *and* make maintenance for the commons lib (which notorously lacks contriubutions and volunteer devs) easier: That's what I personally consider a win-win.

Adding this to the Readme is fine for me - in fact it was there once iirc, but I removed it because it seemed obvious.

But maybe I missed something, so I'm looking forward to your edit describing the drawbacks you are suffering a bit more verbose.

p1gp1g:
> > equals and hashCode aren't used here isn't it ?
> 
> You make assumptions about other third party apps, that may rely on a working comparison or sorting or backpressure detection - all features that rely on a proper `equals` method.

Indeed. And given that even Nextcloud App targets SDK>24, there is no reason to not use those API.

Nevertheless, I think it is better to let the projects using this lib chose if they want to support old SDK and desugare API themselves:
Close in favor of #781

PR #781, opened 2024-11-11
Do not require desugaring

p1gp1g:
Desugaring is used by the library for features available from SDK 24. In other words, desugaring is not needed for applications targeting SDK>=24. This is by the way the minSdk of Nextcloud app since a year [1]

If we remove the compile options from the library gradle, it is possible for applications targetting SDK >= 24 to not care about this property, and for other applications to continue supporting older SDK by applying desugaring by themselves. 

I've added a note to the readme to clarify this requirement.

 [1] https://github.com/nextcloud/android/releases/tag/rc-3.27.0-01

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ This library allows you to use accounts as well as the network stack provided by
 You can check out the [sample app](https://github.com/nextcloud/Android-SingleSignOn/tree/master/sample) which uses this library to fetch some information via SSO from a Nextcloud instance.
 The sample app uses the [Retrofit approach](#51-using-retrofit). Be aware though, that it is for demonstration purposes only. Exception handling, state management etc. must be implemented depending on your use case.
 
+If you need to support __SDK lower than 24__, you must enable [library desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring).
+
 ### 1) Add this library to your project
 
 ```gradle
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -55,7 +55,6 @@ android {
     }
 
     compileOptions {
-        coreLibraryDesugaringEnabled true
         sourceCompatibility JavaVersion.VERSION_17
         targetCompatibility JavaVersion.VERSION_17
     }
@@ -114,8 +113,6 @@ detekt {
 }
 
 dependencies {
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.3'
-
     implementation "androidx.appcompat:appcompat:1.7.0"
     implementation 'androidx.annotation:annotation:1.9.1'
     implementation 'androidx.core:core:1.13.1'


p1gp1g:
Given that Nextcloud App targets SDK 24+, we can also drop support for SDK 21.

tobiasKaminsky:
Makes sense for me.
Thank you for your contribution!

Can you also sign DCO?

stefan-niedermann (commented):
- Let the consuming app decide whether or not it wants to desugar is fine for me
- In this case I assume that the `sample` module required those settings, no?
- I haven't found any hints, that the corelib desugaring library does by definition only provide backports and desugaring for SDK < 24, where did you read that? Actually, it might be the *current* situation, but if newer syntax like Pattern Matching will be supoorted in a future desugaring release, this would affect also Sdk > 24, don't you agree? You even [mentioned](https://github.com/nextcloud/Android-SingleSignOn/pull/779#issue-2648860525) `Streams` and `Optional`s getting available for Sdk >= 24 yourself.

Honestly I still can't see the drawback, it feels like going back to stoneage because - well, why? Even after your edit, I can only see two arguments

> It makes 3rd party app setup harder

3 lines of config not harming anything but optionally allow to use nearly up to date syntax

> Dropping support for Sdk < 24

- Desugarung will bring new features to Sdks > 24 (as they did for Java 8+ fearures and then added Java 11+ features later).
- Noone will stop you from dropping Sdk < 24, even if the language features are desugared.

I am not the maintainer of SSO, but from my side the drawbacks are way larger than the benefits. 👎

p1gp1g:
> * Let the consuming app decide whether or not it wants to desugar is fine for me
> * In this case I assume that the `sample` module required those settings, no?

It has already those settings. That's the point

> * I haven't found any hints, that the corelib desugaring library does by definition only provide backports and desugaring for SDK < 24, where did you read that? Actually, it might be the _current_ situation, but if newer syntax like Pattern Matching will be supported in a future desugaring release, this would affect also Sdk > 24, don't you agree? You even [mentioned](https://github.com/nextcloud/Android-SingleSignOn/pull/779#issue-2648860525) `Streams` and `Optional`s getting available for Sdk >= 24 yourself.

This is indeed the current situation. `Streams` and `Optional` are "backported" to SDK < 24. For instance, if [`Duration`](https://developer.android.com/reference/java/time/Duration) were used, then consuming app would require desugaring if they target SDK < 26. Then the readme can be updated. We can also suggest project to add desugaring no matter their SDK.

> Honestly I still can't see the drawback, it feels like going back to stoneage because - well, why? Even after your edit, I can only see two arguments
> 
> > It makes 3rd party app setup harder
>
> 
> 3 lines of config not harming anything but optionally allow to use nearly up to date syntax
> 
> > Dropping support for Sdk < 24
> 
> * Desugarung will bring new features to Sdks > 24 (as they did for Java 8+ fearures and then added Java 11+ features later).
> 
> * Noone will stop you from dropping Sdk < 24, even if the language features are desugared.
> 
> 
> I am not the maintainer of SSO, but from my side the drawbacks are way larger than the benefits. 👎

I think this is fine if you _need_ it. This is not the case here. So better avoid consumer projects to set up a compile option, tracking versions of a library and adding compilation steps and dependencies. Don't you think ?

p1gp1g:
@tobiasKaminsky I've signed commits

stefan-niedermann:
What I am afraid of is third party apps crashing / throwing errors that won't be detected at build time but at runtime on the users with old devices :-/

p1gp1g:
> > What I am afraid of is third party apps crashing / throwing errors that won't be detected at build time
> 
> Can this happen? I thought this is checked during build time.

This is runtime. But any app using this library have already set desugaring.

It seems this issue is common to some other projects, and it may get fixed by AGP some day: https://issuetracker.google.com/issues/339232491

Some say that the current solution is what is done with this PR, but is indeed fragile https://issuetracker.google.com/issues/341002194#comment14

> If the library might need coreLibraryDesugaring (depending on which part of the library is used, or if building for a high enough minSdk), then maybe not use coreLibraryDesugaring for the library and leave it to the app developers to decide. That will be fragile though, so not sure that is a good path.

stefan-niedermann:
> But any app using this library have already set desugaring.

You mean 

> any app using this library **before this MR** have already set desugaring.

Means also: New 3rd party apps may encounter the mentioned runtime issues (and may not notice it until error reports - for each code part that uses modern Java individually).

Don't get me wrong, this is nothing personal, but from my point of view it is a technically bad decision to drop desugaring because some 3rd party apps *currently*¹ don't *need* it.

¹ as [mentioned above](https://github.com/nextcloud/Android-SingleSignOn/pull/781#pullrequestreview-2428832054): New Java language features may come through desugaring and will provide value even for `minSdk` > 24 in the future.

Maybe some more opinions? @David-Development @AndyScherzinger

tobiasKaminsky:
In general a library should be able to support as many apps/devices/sdk/versions as possible.
If we shrink these supported things with this PR, I prefer to not merge it.

p1gp1g:
I agree, and it doesn't remove support from anything

tobiasKaminsky:
@stefan-niedermann are you fine with it? :+1:

stefan-niedermann:
@tobiasKaminsky I didn't change my mind, so if you mean

>  I prefer to not merge it.

then yes: I would prefer to *not* merge this PR.

tobiasKaminsky:
:+1: then let us not merge it as to above discussed reasons.

@p1gp1g thank you for your ideas and I hope you continue hacking with us :)

PR #1023, opened 2026-07-08
fix(api): don't crash when the service connects after the API was closed

MiMoHo:
# fix(api): don't crash when the service connects after the API was closed

## Problem

Host apps using this library keep crashing with:

```
java.lang.NullPointerException: Attempt to invoke interface method
'void com.nextcloud.android.sso.api.NextcloudAPI$ApiConnectedListener.onConnected()'
on a null object reference
    at com.nextcloud.android.sso.api.AidlNetworkRequest$1.onServiceConnected(AidlNetworkRequest.java:69)
```

Reported for example in nextcloud/notes-android#2622 (labeled "Upstream :: SSO").

## Root cause

`NextcloudAPI` starts binding to the `AccountManagerService` asynchronously.
When the consumer calls `NextcloudAPI#close()` before the connection is
established (e.g. after a request error invalidates a cached API instance),
`NetworkRequest#close()` nulls the `ApiConnectedListener` — but
`AidlNetworkRequest#unbindService()` only called `Context#unbindService()`
when the connection was already established (`mBound == true`). The pending
`ServiceConnection` therefore stayed registered with the framework and later
delivered `onServiceConnected()` on the main thread, which invoked
`onConnected()` on the nulled callback and crashed the host app.

## Fix

- Track whether a binding has been requested and always unbind in that case,
  so a not-yet-established connection is released on `close()` as well.
- Guard `onServiceConnected()` against the remaining race where `close()`
  runs on another thread while the connection callback is already being
  dispatched: capture the callback locally and ignore stale connections.

## Tests

Three new unit tests in `AidlNetworkRequestTest`:

- the regular flow still invokes the callback,
- a stale `onServiceConnected()` after `close()` no longer throws
  (fails with exactly the reported NPE without the fix),
- `close()` before the connection is established unbinds the
  `ServiceConnection` (fails without the fix).

`unitTests.returnDefaultValues = true` was enabled so the plain JVM tests can
run `connect()`, which touches `Intent`/`ComponentName`. All existing tests
and the CI checks (`test`, `assembleDebug`, `lib:lint`, `lib:detekt`,
`lib:spotlessKotlinCheck`) pass locally.

## AI disclosure

This contribution was developed with AI assistance (Claude Code, model
claude-fable-5). I have reviewed, tested and take responsibility for the
changes, in accordance with the Nextcloud AI contribution policy.

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -92,6 +92,7 @@ android {
     }
     testOptions {
         targetSdk 36
+        unitTests.returnDefaultValues = true
     }
 }
 
--- a/lib/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
+++ b/lib/src/main/java/com/nextcloud/android/sso/api/AidlNetworkRequest.java
@@ -48,7 +48,8 @@ public class AidlNetworkRequest extends NetworkRequest {
     private static final String TAG = AidlNetworkRequest.class.getCanonicalName();
 
     private IInputStreamService mService = null;
-    private final AtomicBoolean mBound = new AtomicBoolean(false); // Flag indicating whether we have called bind on the service
+    private final AtomicBoolean mBound = new AtomicBoolean(false); // Flag indicating whether the service connection has been established
+    private final AtomicBoolean mBindingRequested = new AtomicBoolean(false); // Flag indicating whether we have called bind on the service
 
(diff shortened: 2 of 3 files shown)


tobiasKaminsky (approved):

tobiasKaminsky:
Thank you very much!

PR #1033, opened 2026-07-15
feat: Implement ActivityResultContract

### LINKED ISSUE #448: Support Activity Result API
Here is how I integrate NC SSO:

```kotlin
override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
    super.onActivityResult(requestCode, resultCode, data)

    when (resultCode) {
        AppCompatActivity.RESULT_CANCELED -> setButtonsEnabled(true)

        else -> {
            AccountImporter.onActivityResult(
                requestCode,
                resultCode,
                data,
                this,
            ) { onNextcloudAccountAccessGranted(it) }
        }
    }
}
```

It works fine, except for showing deprecation warnings:

> 'onActivityResult(Int, Int, Intent?): Unit' is deprecated. Deprecated in Java

> This method has been deprecated in favor of using the Activity Result API which brings increased type safety via an ActivityResultContract and the prebuilt contracts for common intents available in androidx.activity.result.contract.ActivityResultContracts, provides hooks for testing, and allow receiving results in separate, testable classes independent from your fragment. Use registerForActivityResult(ActivityResultContract, ActivityResultCallback) with the appropriate ActivityResultContract and handling the result in the callback.

It looks like the modern way to get a `SingleSignOnAccount` is to create an `ActivityResultContract<Nothing, SingleSignOnAccount>` or something like that and to ship it with the library. Are there any reasons for not supporting Activity Result API except for the lack of dev time?

stefan-niedermann:
Motivation:
- Provide state-of-the-art access to Nextcloud Single Sign On
- Hide complete permission granting and account importing process behind a clear API surface
- Deprecate methods that rely on deprecated Android framework methods like onActivityResult
- Enhance maintainability and make refactoring easier in the future when the deprecated methods have eventually been removed or marked as private

Refs: https://github.com/nextcloud/Android-SingleSignOn/issues/448

Co-authored-by: Gemini

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccount.java
+++ b/lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccount.java
@@ -1,0 +1,37 @@
+package com.nextcloud.android.sso;
+
+import static com.nextcloud.android.sso.Constants.NEXTCLOUD_SSO;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+
+import androidx.activity.result.contract.ActivityResultContract;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.nextcloud.android.sso.model.SingleSignOnAccount;
+
+/**
+ * {@link ActivityResultContract} to import a Nextcloud account.
+ * The result account may be <code>null</code> if something went wrong or the user canceled the import.
+ *
+ * @see <a href="https://developer.android.com/training/basics/intents/result"><code>ActivityResultContract</code></a>
+ */
+public class ImportSsoAccount extends ActivityResultContract<Void, SingleSignOnAccount> {
+
+    @NonNull
+    @Override
+    public Intent createIntent(@NonNull Context context, Void input) {
+        return new Intent(context, ImportSsoAccountActivity.class);
+    }
+
+    @Override
+    public SingleSignOnAccount parseResult(int resultCode, @Nullable Intent intent) {
+        if (resultCode == Activity.RESULT_OK && intent != null) {
+            return (SingleSignOnAccount) intent.getSerializableExtra(NEXTCLOUD_SSO);
+        }
+
+        return null;
+    }
+}
--- a/lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccount.java
+++ b/lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccount.java
@@ -1,3 +1,9 @@
+/*
+ * Nextcloud Android SingleSignOn Library
+ *
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
 package com.nextcloud.android.sso;
 
 import static com.nextcloud.android.sso.Constants.NEXTCLOUD_SSO;
--- a/lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
+++ b/lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
@@ -1,0 +1,97 @@
+/*
+ * Nextcloud Android SingleSignOn Library
+ *
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package com.nextcloud.android.sso;
+
+import android.content.Intent;
+import android.os.Bundle;
+import android.util.Log;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+import com.nextcloud.android.sso.exceptions.AccountImportCancelledException;
+import com.nextcloud.android.sso.exceptions.AndroidGetAccountsPermissionNotGranted;
+import com.nextcloud.android.sso.exceptions.SSOException;
+import com.nextcloud.android.sso.ui.UiExceptionManager;
+
+/// This is a Trampoline Activity that takes care about requesting all necessary permissions and guiding the user through the complete import flow.
+/// It is extremely important to catch *any* error and finish this activity because otherwise, the Trampoline Activity will stay on top but transparent, so users can't see it and users can't interact with their (visible) 3rd party app anymore.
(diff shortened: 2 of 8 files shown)


stefan-niedermann:
# Additional information

## Exception Handling

A third party app can start the import process and gets either a `SingleSignOnAccount` instance or `null`. Displaying errors to the users is handled intentionally completely within the SSO lib.

This aligns with other `ActivityResultContract`s and gives the SSO lib the full control over exception handling for a uniform user experience across various 3rd party apps.

## Theming

There are multiple sources of dialogs that may look different within the flow:
  - Native android dialogs like the account chooser
  - Permission dialogs from the files app
  - Exception dialogs from the SSO lib

Given this does not affect the (now) deprecated way of importing accounts, this can be enhanced progressively in my opinion, so this should not be a show stopper.
Maybe @AndyScherzinger has some ideas how to unify the styling in the complete flow?

## Testing

Tested successfully manually on AVD API level 24 without Google Play Services and API Level 36.1 with Google Play Services.

AndyScherzinger:
> Given this does not affect the (now) deprecated way of importing accounts, this can be enhanced progressively in my opinion, so this should not be a show stopper.
> Maybe @AndyScherzinger has some ideas how to unify the styling in the complete flow?

@stefan-niedermann Likely depends on how they have been implemented. Lately we migrated various parts of the UI to Composables and with that we get theming "for free" in the sense that we now just to proper design tokens for styling and inject the apps server theme and are done, everything else properly managed by the Composables themselves now. SO no need for calling theme APIs any longer 🚀

lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private static final String TAG = ImportSsoAccountActivity.class.getSimpleName();
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        if (savedInstanceState == null) {
+
+            try {
```
alperozturk96: Why nested try block used here?

lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
```diff
@@ -0,0 +1,97 @@
...
+import androidx.appcompat.app.AppCompatActivity;
+
+import com.nextcloud.android.sso.exceptions.AccountImportCancelledException;
+import com.nextcloud.android.sso.exceptions.AndroidGetAccountsPermissionNotGranted;
+import com.nextcloud.android.sso.exceptions.SSOException;
+import com.nextcloud.android.sso.ui.UiExceptionManager;
+
+/// This is a Trampoline Activity that takes care about requesting all necessary permissions and guiding the user through the complete import flow.
+/// It is extremely important to catch *any* error and finish this activity because otherwise, the Trampoline Activity will stay on top but transparent, so users can't see it and users can't interact with their (visible) 3rd party app anymore.
+public class ImportSsoAccountActivity extends AppCompatActivity {
```
alperozturk96, line 24: Can't we use Kotlin for new files?

lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private static final String TAG = ImportSsoAccountActivity.class.getSimpleName();
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        if (savedInstanceState == null) {
```
alperozturk96: We can fail fast here instead nested if.

lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Nextcloud Android SingleSignOn Library
+ *
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
alperozturk96:
```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccount.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.nextcloud.android.sso.model.SingleSignOnAccount;
+
+/**
+ * {@link ActivityResultContract} to import a Nextcloud account.
+ * The result account may be <code>null</code> if something went wrong or the user canceled the import.
+ *
+ * @see <a href="https://developer.android.com/training/basics/intents/result"><code>ActivityResultContract</code></a>
+ */
+public class ImportSsoAccount extends ActivityResultContract<Void, SingleSignOnAccount> {
```
alperozturk96, line 27: Can't we use Kotlin for new files?

alperozturk96 (changes requested):

lib/src/main/java/com/nextcloud/android/sso/ImportSsoAccountActivity.java
stefan-niedermann: Sure, one can do this. Please note, that (speaking for me personally only) enforcing Kotlin will stop further contributions from my side to this library. I don't want to trigger a "Kotlin vs. Java" discussion, this is a personal decision.

alperozturk96: Of course Java is perfectly fine; it was just a question for new file.

Thank you for the PR :)

stefan-niedermann:
@alperozturk96 I think I addressed all further issues you mentioned :slightly_smiling_face:

alperozturk96:
> @alperozturk96 I think I addressed all further issues you mentioned 🙂

Hey, thanks for the update. Only two small things left :)

Screenshot 2026-08-07 at 15 17 10


and:

Screenshot 2026-08-07 at 15 18 26

Thank you.

stefan-niedermann:
Done :+1:

alperozturk96 (approved):

